PR #324, opened 2014-07-19
Remove MultiJSON dependency

bf4:
This is stepwise

0. Remove the MultiJSON gem dependency.  (Though the tests still get it via shoulda via active_support).
1. Show that the code with the adapters doesn't break multi-json (which is removed and yet still available)
2. Remove Appraisal and its generated gemfiles, and in travis.yml
3. Make JSON gem the default parser (and only test it), but allow for continued use of MultiJSON or any JSON parser that has the load/dump interface.


sferik:
Resolved by https://github.com/colszowka/simplecov/commit/ba6af083b223fcb2c25b8295abbae0db8a2f7696.

PR #331, opened 2014-09-14
Filter `/vendor/bundle` by default

andyw8:
The Rails profile filters out `/vendor/bundle`, but this is also needed for many other kinds of Ruby projects.

A GitHub search illustrates how often developers need to configure this:

https://github.com/search?utf8=%E2%9C%93&q=simplecov+%22add_filter*vendor%22&type=Code&ref=searchresults

I can't think of any downside of having this as the default behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,7 +333,7 @@ that shows how you can achieve just that:
 
 ```ruby
 SimpleCov.start :rails do
-  filters.clear # This will remove the :root_filter that comes via simplecov's defaults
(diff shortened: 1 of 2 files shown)


andyw8:
@colszowka @bf4 any thoughts on this?

sferik:
:+1: Merging.

nTraum:
Niiiice. :+1:

PR #332, opened 2014-09-17
Added codecov.io custom reporter

stevepeak:
Hey friends! We are a startup in Beta and have build a connector using this great product called simplecov :+1: 

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -557,6 +557,11 @@ SimpleCov.formatters = [
 Apart from the direct companion [simplecov-html], there are other formatters
 available:
 
+#### [codecov](https://github.com/codecov/codecov-ruby)
+*by [Codecov](https://codecov.io/)*
+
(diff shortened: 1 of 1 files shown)


colszowka:
@stevepeak Merged and moved to a separate, new document alongside some other hosted services: https://github.com/colszowka/simplecov/blob/master/doc/commercial-services.md

@bf4 Let's keep this within the source code. I created a new doc/ dir and moved the open source formatters there too, with a reference from the main readme. Maybe we can move out some more specific problem solutions from the readme over there, too. 

Why the doc dir? This way we can bundle the docs with the gem package, and also I think it's easier to maintain the normal PR -> Merge flow for edits instead of having another thing to watch and review by having the Wiki enabled.

stevepeak:
Thanks you :)

PR #341, opened 2014-10-24
Add writeup about using with Spring to README

swrobel:


colszowka:
@swrobel Thanks, merged!

PR #353, opened 2014-12-07
Fix mispointed link in CHANGELOG.md

dleve123:
I hit the `at_exit` bug when running my suite on Travis CI with the [`codeclimate-test-reporter `](https://github.com/codeclimate/ruby-test-reporter) that (I think) manifested from the bug fixed in https://github.com/colszowka/simplecov/pull/302. The CHANGELOG as it currently stands made me think that the issue was never resolved as it pointed to #303! Pointing to the correct PR should appease/help desperate build fixers everywhere :smiley:!


bf4:
:rainbow:

PR #356, opened 2014-12-31
Improve command name docs

gtd:
It actually took me several hours to figure out why I wasn't getting the coverage I expected, granted that's because my test suite is slow and I'm new to SimpleCov, but I think this additional paragraph will save someone some time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,6 +403,11 @@ SimpleCov.command_name "features"
 Note that this only has to be invoked ONCE PER TEST SUITE, so even if you have 200 unit test files, specifying it in
 some_test.rb is enough.
 
+Last but not least **if multiple suites resolve to the same `command_name`** be aware that the coverage results **will
+clobber each other instead of being merged**.  SimpleCov is smart enough to detect unique names for the most common
+setups, but if you have more than one test suite that doesn't follow a common pattern then you will want to manually
+ensure that each suite gets a unique `command_name`.
+
 [simplecov-html] prints the used test suites in the footer of the generated coverage report.
 
 ### Timeout for merge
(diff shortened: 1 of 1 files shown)


bf4:
Looks good. Would you mind adding some of the info from https://github.com/colszowka/simplecov/issues/340 about environmental variables?  Let me know if you'd like any help adding it, or if you'd just like me to do it.

Quoted below for convenience.

> Parallel testing

> Using .simplecov rather than separately requiring SimpleCov multiple times is recommended if you are merging multiple test frameworks like Cucumber and RSpec that rely on each other, as invoking SimpleCov multiple times can cause coverage information to be lost.

> You'll also want to use multiple 'command_names' to differentiate reports being merged in together.

> See https://github.com/colszowka/simplecov/blob/master/lib/simplecov/command_guesser.rb#L19 that sets an unset command_name from the env if ENV['PARALLEL_TEST_GROUPS'] && ENV['TEST_ENV_NUMBER'], or from the command that ran the tests, or from any defined constants. to be sure, you may want to set it yourself, e.g. command_name "rails_app_#{$$}" # $$ is the processid

> And then add something like merge_timeout 3600 # 1 hour should cover how long it takes the tests to run

gtd:
Okay, I added a paragraph about that and rebased.  Let me know what you think.

bf4:
@colszowka @envygeeks @sferik looks good to me. Any comments?

sferik:
Very nice improvements to the `README`. Thanks!

colszowka:
:+1: Thanks!

bf4:
@colszowka Could you add @envygeeks as a collab?  Also, any thoughts about moving `simplecov` to an org?

PR #374, opened 2015-02-23
Protect from reading 'null' from file

avsej:


bf4:
Should this maybe be addressed elsewhere to that such a file doesn't exist?

sferik:
This should no longer be necessary after replacing `multi_json`, which returns `nil` when parsing a `null` string, with `json`, which raises:

    JSON::ParserError: 757: unexpected token at 'null'

PR #376, opened 2015-03-07
Fixes order dependencies in unit tests

### LINKED ISSUE #362: Fix order dependencies in unit tests
After [switching from Test::Unit to Minitest](https://github.com/colszowka/simplecov/commit/667c3991a16c541609fbeff5a46b92b9505881a9), I started noticing some intermittent build failures. These will typically self-correct by restarting the build. They seem to be caused by an order dependency.

Minitest will force the tests to run in alphabetical order if you add the flag `i_suck_and_my_tests_are_order_dependent!` but I think it would be better to just fix the order dependency. :blush:

There’s a project called [minitest-bisect](https://github.com/seattlerb/minitest-bisect) that should make it easier to track down random failures.

hugopeixoto:
Some tests were changing global configuration settings
without restoring them afterwards. Fixes #362

### DIFF
--- a/test/test_filters.rb
+++ b/test/test_filters.rb
@@ -37,8 +37,12 @@ class TestFilters < Minitest::Test
 
   context "with no filters set up and a basic source file in an array" do
     setup do
-      SimpleCov.filters = []
+      @prev_filters, SimpleCov.filters = SimpleCov.filters, []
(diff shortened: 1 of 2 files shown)


colszowka:
Excellent, thank you sir!

PR #377, opened 2015-03-12
Only run the at_exit behaviors if the current PID matches the PID that called SimpleCov.start

### LINKED ISSUE #371: Default code path doesn't appear to handle fork correctly?
I had to do this.

https://gist.github.com/clicube/5017378

 This removed what appeared to be superfluous output:

    Coverage report generated for RSpec to /home/samuel/Documents/Programming/rubydns/coverage. 0 / 599 LOC (0.0%) covered.

multiple times per run.

Perhaps this should be the default behaviour?

coderanger:
This handles the case where a test calls fork at some point, the new process shouldn't try to run coverage formatters or check the minimum coverage level.

This resolves #371.

### DIFF
--- a/features/rspec_basic.feature
+++ b/features/rspec_basic.feature
@@ -14,14 +14,15 @@ Feature:
     When I open the coverage report generated with `bundle exec rspec spec`
     Then I should see the groups:
       | name      | coverage | files |
-      | All Files | 91.23%   | 6     |
+      | All Files | 91.8%   | 7     |
 
     And I should see the source files:
       | name                                    | coverage |
       | lib/faked_project.rb                    | 100.0 %  |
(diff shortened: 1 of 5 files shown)


coderanger:
Not sure whats up with that test failure, can't repro it locally.

colszowka:
Great, thanks. Don't worry about the failing tests, we had an order issue there, see #362 - which should be resolved now that I've merged #376.

grosser:
FYI this breaks using forking test runners with code coverage for example https://github.com/grosser/codeclimate_batch + https://github.com/grosser/forking_test_runner

not a big issue, just wanted you to be aware that this is a usecase / tell me if this looks wrong and I should be doing something else

```Ruby
fork do
  SimpleCov.pid = Process.pid
  ... run test ...
end
```

Issue #381, opened 2015-03-30
How to configure SimpleCov with Spring?


onebree:
**Rails 4.2.0, RSpec 3.1, SimpleCov 0.9.2**

I have the following in my `rails_helper.rb` file, _outside_ of the `RSpec.configure` block
```ruby
require 'simplecov'
SimpleCov.start 'rails' do
  add_filter '/app/models/legacy/'
  add_filter '/app/controllers/playlists_controller_legacy.rb'
  add_filter '/app/models/login.rb'
  add_filter '/app/models/login_auth.rb'
  add_filter '/app/models/login_config.rb'
end
```

I have read online that some people find running `spring rspec` returns high or even 100% coverage. It does not seem that way for our app, but I would just like to make sure. The LOC count seems a little low (short by 1000 maybe?), although again that could be from recent refactoring I have not viewed the commits for.

And my `config/spring.rb` file contains nothing about SimpleCov. Would I put the above configuration in the spring file itself, or under the `Spring.after_fork` block?

MasterSlowPoke:
I definitely have an issue with spring and simplecov:

```bash
$ bundle exec rspec
Finished in 15.98 seconds (files took 8.6 seconds to load)
87 examples, 0 failures, 5 pending

Coverage report generated for RSpec to /Users/craigsniffen/Documents/Projects/71lbs/coverage. 2372 / 7486 LOC (31.69%) covered.

$ bundle exec spring rspec
Finished in 13.9 seconds (files took 0.65848 seconds to load)
87 examples, 0 failures, 5 pending

Coverage report generated for RSpec to /Users/craigsniffen/Documents/Projects/71lbs/coverage. 18 / 27 LOC (66.67%) covered.
```

This is with ```config.eager_load = true``` set in the test environment, as I'd like to have all the files, even untested ones, in the coverage report. Setting this to false 'fixes' spring:

```bash
$ bundle exec rspec
Finished in 20.97 seconds (files took 11.15 seconds to load)
87 examples, 0 failures, 5 pending

Coverage report generated for RSpec to /Users/craigsniffen/Documents/Projects/71lbs/coverage. 1505 / 3530 LOC (42.63%) covered.

$ bundle exec spring rspec
Finished in 15.01 seconds (files took 0.89234 seconds to load)
87 examples, 0 failures, 5 pending

[...]

PR #382, opened 2015-04-02
update README with Spring config

onebree:
I hope the way I phrased everything is okay. As mentioned in issue #381 I set two configurations to `false` in my environment files. Such settings were also recommended by Rails. And although Spring preloads your environment, I find that still disabling `eager_load` helps Spring cooperate with SimpleCov

**ALSO** I know the contributing doc says to write tests for any changes, but since this is just the README, I do not think it is necessary. I can provide my own stats if needed (running my own suite).

```
DEPRECATION WARNING: The configuration option `config.serve_static_assets` has been renamed to
`config.serve_static_files` to clarify its role (it merely enables serving everything in the
`public` folder and is unrelated to the asset pipeline). The `serve_static_assets` alias will be
removed in Rails 5.0. Please migrate your configuration files accordingly. (called from block in
<top (required)> at /var/www/rails/hosted-fork/hpbxgui/config/environments/development.rb:15)

config.eager_load is set to nil. Please update your config/environments/*.rb files accordingly:

  * development - set it to false
  * test - set it to false (unless you use a tool that preloads your test environment)
  * production - set it to true
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -604,6 +604,8 @@ if ENV['RAILS_ENV'] == 'test'
 end
 ```
 
+(3) Run `spring rspec <path>` as normal. Remember to run `spring stop` after making important changes to your app or its specs!
+
 ## Contributing
 
 See the [contributing guide](https://github.com/colszowka/simplecov/blob/master/CONTRIBUTING.md).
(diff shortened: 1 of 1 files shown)


sferik:
Thanks for this update!

PR #384, opened 2015-04-21
Adds coverage calculation by individual file

ptashman:
This allows you to calculate code coverage by each individual file. It also allows you to set a separate minimum requirement for this metric. It will fail the build if minimum coverage in the whole app is met but minimum coverage in one or more files are not met. This request also includes tests for these features.


ptashman:
Any input on this?

ptashman:
Are there any improvements needed for this to be pulled in?

lib/simplecov/file_list.rb
```diff
@@ -26,6 +26,13 @@ def skipped_lines
       map { |f| f.skipped_lines.count }.inject(&:+)
     end
 
+    # Computes the coverage based upon lines covered and lines missed for each file
+    # Returns an array with all coverage percentages
+    def covered_percentages
+      return [] if empty?
```
xaviershay: this is redundant, right?

lib/simplecov/defaults.rb
```diff
@@ -61,11 +61,15 @@
 
   if SimpleCov.result? # Result has been computed
     covered_percent = SimpleCov.result.covered_percent.round(2)
+    covered_percentages = SimpleCov.result.covered_percentages.map { |p| p.round(2) }
 
     if @exit_status == SimpleCov::ExitCodes::SUCCESS # No other errors
       if covered_percent < SimpleCov.minimum_coverage # rubocop:disable Metrics/BlockNesting
         $stderr.printf("Coverage (%.2f%%) is below the expected minimum coverage (%.2f%%).\n", covered_percent, SimpleCov.minimum_coverage)
         @exit_status = SimpleCov::ExitCodes::MINIMUM_COVERAGE
+      elsif covered_percentages.any? { |p| p < SimpleCov.minimum_coverage_by_file } # rubocop:disable Metrics/BlockNesting
```
xaviershay: is there a way to include the offending file name here?

xaviershay:
Sorry this was ignored. We all must have assumed someone else was going to look at it :(

I see some tests in commits, but not in the resulting diff. Maybe something to do with the merge ... can you rebase it?

Feature-wise: is there a particularly compelling reason for doing this rather than just overall percentage? In practice does it result in different behaviour? I'm not necessarily against it, I just want to have good rationale for expanding the API surface area. In my mind, more configuration options is a bad thing by default so we need to have really good reasons for adding them.

Thank you for your contribution!

ptashman:
This is a feature my company wanted. If you have a minimum requirement of X for the whole system, then department 1 could slack off with only X-20% coverage while department 2 has to make up for their lack with X+20%. In this case, department 1's code won't be covered enough and department 2 will be writing unnecessary tests just to get the whole system up to the benchmark so they can deploy. This minimum coverage by file requirement prevents such a scenario from occurring. All test files must be appropriately covered. If a user does not need this feature, they can simply ignore it.

bf4:
is the tl;dr of this that a minimum_coverage of 80% could mean some files are 100% and others are 50% etc., and this PR would say that if minimum_coverage_by_file is 80%, then the 50% files would fail the build.  I haven't looked at the code carefully, but I image this could be used in conjunction with minimum_coverage so that the pseudo code is 

```ruby
if average_coverage < minimum_coverage
  fail_build
elsif coverage_per_file.any? {|coverage, file| coverage < minimimum_coverage_by_file }
  fail_build
end
```

ptashman:
Yes bf4. That's exactly the reason for this.

lib/simplecov/configuration.rb
```diff
@@ -7,7 +7,7 @@
 # SimpleCov documentation for further info.
 #
 module SimpleCov
-  module Configuration
+  module Configuration # rubocop:disable all
```
xaviershay: seems kinda broad?

xaviershay: or wait is it just for this line? Maybe deserves explanatory comment .. don't want this being copied everywhere.

xaviershay:
Looking at rubocop failures ... doesn't seem a great reason to be ignoring them beyond personal preference (i.e. no parallel assignment). Whole point of rubocop is to avoid bike-shedding over personal preference, so I'd prefer the disables removed.

Length of config module warning is maybe ok to disable.

xaviershay:
seems good as a feature though

xaviershay:
oh should add some docs to the README as well. There's already a "Minimum Coverage" section, so could just add there.

ptashman:
I believe it's ok to merge now. Let me know if you have any more suggestions.

xaviershay:
Looks good, let me get master green later today then I'll rebase/squash this down to a single commit and merge it.

xaviershay:
Superceeded by https://github.com/colszowka/simplecov/pull/392

PR #385, opened 2015-04-27
Require the Ruby English library prior to use

nbibler:
The use of English's `$ERROR_INFO`, rather than the standard `$!`, causes a warning in Ruby when no exceptions have previously occurred. Requiring the English library preloads and initializes the global and removes this warning.

### DIFF
--- a/lib/simplecov/defaults.rb
+++ b/lib/simplecov/defaults.rb
@@ -1,3 +1,5 @@
+require "English"
+
 # Load default formatter gem
 require "simplecov-html"
 


colszowka:
This is literally the first time I hear about such a thing. Anyways, the build failed, and as I [managed to dig up](http://docs.ruby-lang.org/ja/1.9.3/library/English.html) you have to actually require a capital `English`. Can you please update and send it in again? Also, is there some way we might be able to test this?

nbibler:
Wow, good catch! I think this might be the only standard library in Ruby that I've seen that is required with a titlecased name.

As for testing it, _maybe_ there's something that can be done with a variable defined check, but I kind of doubt it...

nbibler:
And, of course all of the tests pass locally... because MacOS X is case-insensitive by default. :frowning: 

I'll force-push a new commit onto my branch and this pull request in a moment.

nbibler:
I attempted to update the code to make the Ruby warnings more apparent, but I'm not seeing the warnings when running the local test suite.

This is the `Rakefile` modification if you'd like to enable Ruby warnings:

```diff
diff --git a/Rakefile b/Rakefile
index fd64fc3..5a92f24 100644
--- a/Rakefile
+++ b/Rakefile
@@ -14,6 +14,7 @@ Rake::Task[:build].prerequisites.unshift :fix_permissions
 
 require "rspec/core/rake_task"
 RSpec::Core::RakeTask.new(:spec) do |spec|
+  spec.ruby_opts = "-w"
   spec.pattern = FileList["spec/*_spec.rb"]
 end
```

calebmeyer:
+1 Using rspec 3.2.0 (no rake task, just the `rspec` command) with simplecov raises this warning for me.

I'm using the default generated spec_helper (from `rspec --init`) with the one change that I require and start simplecov at the top of my spec helper.

bf4:
@colszowka this looks fine to me to merge.

re adding warnings, the problem is that by setting the `RUBY_OPTS=-w` you'll get a warnings from anything in ruby, so we'd need to filter by SimpleCover.root. See https://github.com/metricfu/metric_fu/blob/master/spec/capture_warnings.rb for a way I've done this

xaviershay:
We fixed this a different way, by just not using `ERROR_INFO`.

PR #396, opened 2015-07-06
speed up root_filter

raszi:
```
         start_with?   103.000  i/100ms
               match    14.000  i/100ms
-------------------------------------------------
         start_with?      1.043k (± 3.9%) i/s -      5.253k
               match    147.419  (± 3.4%) i/s -    742.000

Comparison:
         start_with?:     1042.7 i/s
               match:      147.4 i/s - 7.07x slower
```

### DIFF
--- a/lib/simplecov/defaults.rb
+++ b/lib/simplecov/defaults.rb
@@ -3,8 +3,9 @@ require "simplecov-html"
 
 SimpleCov.profiles.define "root_filter" do
   # Exclude all files outside of simplecov root
+  root_filter = /\A#{Regexp.escape(SimpleCov.root)}/io
   add_filter do |src|
-    !(src.filename =~ /^#{Regexp.escape(SimpleCov.root)}/i)
+    !(src.filename =~ root_filter)
   end
 end
 
--- a/lib/simplecov/defaults.rb
+++ b/lib/simplecov/defaults.rb
@@ -4,7 +4,7 @@ require "simplecov-html"
 SimpleCov.profiles.define "root_filter" do
   # Exclude all files outside of simplecov root
   add_filter do |src|
-    !src.filename.downcase.start_with?(SimpleCov.root.downcase)
+    !(src.filename =~ /^#{Regexp.escape(SimpleCov.root)}/i)
   end
 end
 
--- a/lib/simplecov/defaults.rb
+++ b/lib/simplecov/defaults.rb
@@ -4,7 +4,7 @@ require "simplecov-html"
 SimpleCov.profiles.define "root_filter" do
   # Exclude all files outside of simplecov root
   add_filter do |src|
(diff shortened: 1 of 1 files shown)


lib/simplecov/defaults.rb
```diff
@@ -4,7 +4,7 @@
 SimpleCov.profiles.define "root_filter" do
   # Exclude all files outside of simplecov root
   add_filter do |src|
-    !(src.filename =~ /^#{Regexp.escape(SimpleCov.root)}/i)
+    !src.filename.downcase.start_with?(SimpleCov.root.downcase)
```
bf4: I'm trying to think if there's any chance src.filename might be nil

Also, can you try benchmarking with `/^#{Regexp.escape(SimpleCov.root)}/io` (the o means it will only compile the interpolated value once), and also maybe `\A` instead of `^`, and also maybe `ROOT_FILTER = /^{Regexp.escape(SimpleCov.root)}/i)`?

raszi: I am a bit surprised, regex solution is faster, but not because of the extraction and the `o` modifier but because of the `\A` change (which of course makes sense), but I did not expect that it will be faster than `String#start_with?`.

```
         start_with?   105.000  i/100ms
     match_optimized    14.000  i/100ms
   match_optimized_A   102.435k i/100ms
               match    13.000  i/100ms
-------------------------------------------------
         start_with?      1.110k (± 3.5%) i/s -      5.565k
     match_optimized    147.382  (± 6.1%) i/s -    742.000
   match_optimized_A      2.477M (± 6.6%) i/s -     12.395M
               match    143.284  (± 6.3%) i/s -    715.000

Comparison:
   match_optimized_A:  2477292.0 i/s
         start_with?:     1110.0 i/s - 2231.75x slower
     match_optimized:      147.4 i/s - 16808.64x slower
               match:      143.3 i/s - 17289.33x slower
```

bf4: blog post?  That's pretty interesting

bf4: mind gisting your benchmark here? or maybe adding it to `bench/`

bf4:
Which rubies have you tested this against? Only MRI? I wonder if that's an artifact of the regex engine (http://www.geocities.jp/kosako3/oniguruma/ ) ref: discussion https://github.com/colszowka/simplecov/pull/396#discussion_r34044864

raszi:
I used 2.1.5 for the benchmarks.

lib/simplecov/defaults.rb
raszi: Sure: https://gist.github.com/raszi/de8e9efa7f26c6ae729d

bf4:
@colszowka @xaviershay @sferik @amatsuda  LGTM

xaviershay:
LGTM

PR #400, opened 2015-07-21
"warning: global variable `$ERROR_INFO' not initialized"

amatsuda:
This eliminates a Ruby :warning: when run with `RUBYOPT="-w"`.

You can easily reproduce the warning as follows:
`% ruby -w -rsimplecov -e 'SimpleCov.pid = Process.pid'`

### DIFF
--- a/lib/simplecov/defaults.rb
+++ b/lib/simplecov/defaults.rb
@@ -48,11 +48,11 @@ at_exit do
   # If we are in a different process than called start, don't interfere.
   next if SimpleCov.pid != Process.pid
 
-  if $ERROR_INFO # was an exception thrown?
+  if $! # was an exception thrown?
     # if it was a SystemExit, use the accompanying status
     # otherwise set a non-zero status representing termination by some other exception
     # (see github issue 41)
-    @exit_status = $ERROR_INFO.is_a?(SystemExit) ? $ERROR_INFO.status : SimpleCov::ExitCodes::EXCEPTION
+    @exit_status = $!.is_a?(SystemExit) ? $!.status : SimpleCov::ExitCodes::EXCEPTION
   else
     # Store the exit status of the test run since it goes away after calling the at_exit proc...
     @exit_status = SimpleCov::ExitCodes::SUCCESS
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -57,6 +57,7 @@ Style/SpecialGlobalVars:
   Exclude:
     - 'spec/deleted_source_spec.rb'
     - 'spec/return_codes_spec.rb'
+    - 'lib/simplecov/defaults.rb'
 
 Style/StringLiterals:
   EnforcedStyle: double_quotes


xaviershay:
Where is this defined in the first place? Makes me nervous that we're papering over a bigger issue...

amatsuda:
Indeed. I'm sorry that I didn't care about that...

I found that `$ERROR_INFO` was defined [here in the stdlib English.rb](https://github.com/ruby/ruby/blob/fd70f7c4ca79c288976d1c176b23db26d89000f0/lib/English.rb#L51) as just an alias to `$!`.

```
% ruby -e '1/0 rescue p $ERROR_INFO'
nil

% ruby -rEnglish -e '1/0 rescue p $ERROR_INFO'
#<ZeroDivisionError: divided by 0>
```

Updated the code to check `$!`. I think the code looks OK with the existing comment.

xaviershay:
LGTM, not sure why the build is failing :(

amatsuda:
Haven't tracked down deeper yet, but aruba 0.8.x seems broken.
Bundling down aruba to 0.7.4 fixes the problem on my local environment.

lib/simplecov/defaults.rb
```diff
@@ -47,7 +47,7 @@
   # If we are in a different process than called start, don't interfere.
   next if SimpleCov.pid != Process.pid
 
-  if $ERROR_INFO # was an exception thrown?
+  if $! # was an exception thrown?
```
bf4: Perhaps we should instead `require 'English'` at the top of the file?

craiglittle:
Hey there! I stopped by to offer a similar patch.

As @bf4 suggested, it seems like the better fix would be to `require 'english'` instead of assuming it's been done by another library, which will allow us to keep the readability of the English alias.

As @amatsuda pointed out, if we want to get the build running as soon as possible, locking `aruba` to the `v0.7.x` series will at least get the build running again, and then debugging to get on the `v0.8.x` series could be done as a separate step.

amatsuda:
@bf4 @craiglittle Thank you for the suggestion, but to me `$ERROR_INFO` doesn't really add no more "readability" over `# was an exception thrown?`.
Requiring 'English' would add 25 global variables. I don't want to see such a "side effect" just in order to check `$!` value.
`$!` is a relatively popular one among these Perl-ish magical gvars, so I think it's fine here with the comment.

xaviershay:
My two cents: `$ERROR_INFO` is _more_ confusing. It's obvious that `$!` is a special ruby global so is populated by magic, the comment explains. `$ERROR_INFO` feels like it was defined by us somewhere.

craiglittle:
Fair enough, works for me!

@amatsuda Were you going to get the build green by locking down to the `v0.7.x` series of `aruba`?

bf4:
I'll merge once this is green. Thanks @amatsuda ! (and everyone else commenting)

craiglittle:
@amatsuda Looks like there's a different issue now, eh?

amatsuda:
@craiglittle Yeah, this is weird. I cannot reproduce the failure on my local machine... :worried:

amatsuda:
WTF

lib/simplecov/defaults.rb
bf4: I'm working on another app, and it was sufficient to do

```ruby
$ERROR_INFO ||= $!
if $ERROR_INFO.....
```

bf4:
I'm working on another app and got this failure, and it was sufficient to do

```ruby
$ERROR_INFO ||= $!
if $ERROR_INFO.....
```

This is important because it is also used below

` @exit_status = $ERROR_INFO.is_a?(SystemExit) ? $ERROR_INFO.status : SimpleCov::ExitCodes::EXCEPTION`

Would that make people happy?

lib/simplecov/defaults.rb
amatsuda: @bf4 Thank you for the hint! I wasn't aware of the other `$ERROR_INFO` here right after the first emergence :astonished:
And changing all `$ERROR_INFO`s to `$!` made the build green finally!

bf4: great, so rebase and merge?

I have a patch in my library now to add the alias :(

bf4:
Alright! Let's update the CHANGELOG and we can release 0.10.1!

craiglittle:
We're so close to the finish line. Looking forward to getting this warning out of my build!

kdawgwilk:
:+1: to see this released!

bf4:
Would someone mind helping update the CHANGELOG? I am elsewhere

ShamoX:
Hi,

I'm having back this warning :

```shell
$ bundle exec rspec
....
Randomized with seed 8732

*****/vendor/bundle/ruby/2.5.0/gems/simplecov-0.16.1/lib/simplecov.rb:177: warning: global variable `$ERROR_INFO' not initialized
Coverage report generated for RSpec ******
```

It seems that the English library doesn't declare it (I guess I'm wrong, but: https://github.com/rubyworks/english/search?utf8=✓&q=%24ERROR_INFO&type=).

I happens to me since I've included [dry-validation-matcher](url).
But I checked in a clean environment:
```ruby
$ bundle exec pry
[1] pry(main)> require 'english'
=> true
[2] pry(main)> defined? $ERROR_INFO
=> nil
[3] pry(main)> defined? $!
=> "global-variable"
[4] pry(main)>
```

## Fix proposal

lib/simplecov.rb:177
`@exit_exception = defined?($ERROR_INFO) ? $ERROR_INFO : $!`

bf4:
@ShamoX This was fixed nearly three years ago. Either you're on an old version or this is a new issue. 'English' is part of the Ruby standard library.

PR #407, opened 2015-08-17
Avoid bundling aruba 0.8.x for now

amatsuda:
As discussed at #400, simplecov features are currently broken, and that prevents some PRs unrelated to the failing features to get merged in.

This was caused by aruba's internal refactoring between 0.7 and 0.8 around [here](https://github.com/cucumber/aruba/blob/v0.8.1/lib/aruba/cucumber.rb#L339).
It seems like that `all be_...` is expected to to invoke RSpec's built in All matcher, but it actually calls Capybara's `page.all` in our context.

I would raise an issue for the aruba repo later, but let us lock the gem's version here until the problem is fixed there.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,7 @@ group :test do
     gem "i18n", "~> 0.6.11"
   end
   platform :jruby, :ruby_19, :ruby_20, :ruby_21, :ruby_22 do
-    gem "aruba", "~> 0.6"
+    gem "aruba", "~> 0.7.4"
     gem "capybara", "~> 2.4"
     gem "cucumber", "~> 2.0"
     gem "phantomjs", "~> 1.9"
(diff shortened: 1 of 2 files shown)


colszowka:
Interesting that the build is failing due to rubocop, I don't think it's complaint (`replace gsub with tr`) is valid - see https://github.com/bbatsov/rubocop/issues/2061

amatsuda:
Ha! Stupid cop!
But, yeah, that `gsub` can actually be replaced with `tr` in this particular case, as it claims.
Should I better add that commit here to fix the build?

colszowka:
Ah right, that's true - I got confused :) Yes, maybe let's actually do the replace here then :)

amatsuda:
@colszowka Done! The build is :green_heart: now :)

colszowka:
@amatsuda Great, thanks! By the way, I've seen you somewhat active here recently, would you like to be added as a collaborator to the repo?

amatsuda:
@colszowka Ah, thank you. Well... yes. Yes, please.

colszowka:
@amatsuda Great, welcome on board then!

/cc @xaviershay @bf4 @sferik

sferik:
:wave: @amatsuda Welcome to the team!

Issue #406, opened 2015-08-14
Undefined method each for Nil in result_merger


JanStevens:
Hello,

Using latest ruby/simplecov/jenkins.
On our jenkins build system we sometimes see the following error returning when all specs passed and simplecov starts building its report:

```
Coverage report generated for  to /var/lib/jenkins/jobs/SpaceCII/workspace/coverage. 16478 / 19835 LOC (83.08%) covered.
/var/lib/jenkins/jobs/SpaceCI/workspace/vendor/bundle/ruby/2.2.0/gems/simplecov-0.10.0/lib/simplecov/result_merger.rb:47:in `results': undefined method `each' for nil:NilClass (NoMethodError)
    from /var/lib/jenkins/jobs/SpaceCI/workspace/vendor/bundle/ruby/2.2.0/gems/simplecov-0.10.0/lib/simplecov/result_merger.rb:69:in `merged_result'
    from /var/lib/jenkins/jobs/SpaceCI/workspace/vendor/bundle/ruby/2.2.0/gems/simplecov-0.10.0/lib/simplecov.rb:61:in `result'
    from /var/lib/jenkins/jobs/SpaceCI/workspace/vendor/bundle/ruby/2.2.0/gems/simplecov-0.10.0/lib/simplecov/configuration.rb:159:in `block in at_exit'
    from /var/lib/jenkins/jobs/SpaceCI/workspace/vendor/bundle/ruby/2.2.0/gems/simplecov-0.10.0/lib/simplecov/defaults.rb:60:in `call'
    from /var/lib/jenkins/jobs/SpaceCI/workspace/vendor/bundle/ruby/2.2.0/gems/simplecov-0.10.0/lib/simplecov/defaults.rb:60:in `block in <top (required)>'
```

Looking at the code I see that `resultset` can return `nil` when the `stored_data` returns `nil`. For some reason the data is not written yet to the file so simplecov fails. I find this failing a bit ugly. Would it be more preferred to always return a Hash, even when the reading of JSON data fails?

In [result_merger.rb](https://github.com/colszowka/simplecov/blob/master/lib%2Fsimplecov%2Fresult_merger.rb?ts=2#L21)

```
# Loads the cached resultset from JSON and returns it as a Hash
      def resultset
        if stored_data
          begin
            JSON.parse(stored_data) || {}
          rescue
            {}
          end
        else
          {}
        end
      end
```

[...]

PR #408, opened 2015-08-18
Check if stored_data is not empty before parsing it to json

JanStevens:
In light of issue #406

### DIFF
--- a/lib/simplecov/result_merger.rb
+++ b/lib/simplecov/result_merger.rb
@@ -19,9 +19,9 @@ module SimpleCov
 
       # Loads the cached resultset from JSON and returns it as a Hash
       def resultset
-        if stored_data
+        if stored_data && !stored_data.empty?
           begin
-            JSON.parse(stored_data) || {}
+            JSON.parse(stored_data)
           rescue
             {}
           end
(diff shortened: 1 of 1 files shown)


bf4:
Is it possible that the bug is actually that stored_data should be `nil` not `''` or `"null"` and that is a bug originated in how stored_data is generated?  That said, the fix could also be `if stored_data && !stored_data.empty?`

JanStevens:
This largely depends on your JSON implementation, It seems that Oj for example returns nil in case `stored_data` is nil. Standaard JSON lib will raise a `TypeError`

I agree that the fix could also be to check if `stored_data` is not empty (which would make sense)

Regards

bf4:
Would you mind rebasing your commits and adding a test for this?

JanStevens:
Looking at the test suite and current solutions I would say its already being tested:

- https://github.com/colszowka/simplecov/blob/master/spec%2Fmerge_helpers_spec.rb?ts=2#L48-L58

I can add specific test to see that we always return an empty hash incase the file couldn't be read but the `be_empty` seems to be sufficient.

JanStevens:
Lets merge merge merge? xD

xaviershay:
If it was already being tested, then why weren't those tests failing?

xaviershay:
I'm not convinced this addresses the issue. See comments on #406. Closing until someone convinces me otherwise.

PR #421, opened 2015-10-24
Fixes deprecation warning in cucumber related to in_current_dir

hugopeixoto:
I was running tests and noticed there was a deprecation warning, as seen here:

https://travis-ci.org/colszowka/simplecov/jobs/78887751

Not sure if this will work in all the rubies, though. Let's see what travis has to say about that.


xaviershay:
Still seems good, merging.

Issue #16, opened 2010-11-15
Files with no coverage


rkistner:
Files without any coverage are not shown at all. It is important for me to be able to view these files.

As this is not detected by the built-in Coverage module, perhaps an option could be added to search for files in the current folder?

I'll try to create a proof-of-concept for this.

adampope:
We encountered a problem with this method of tracking all files in #394 as it was marking lines as uncovered that shouldn't be.  So we went looking for another solution and we found that adding `Rails.application.eager_load!` after `SimpleCov.start` in `test_helper.rb` and making sure that our lib files were in eager_load_paths in `application.rb`

```
config.eager_load_paths += %W(#{config.root}/lib)
config.eager_load_paths += Dir["#{config.root}/lib/**/"]
```

solved all of our problems.  All files are being picked up for coverage purposes and we are not getting lines that should be ignored marked as uncovered.

xaviershay:
See the `track_files` option added in #422 above. Will be released in 0.11.

PR #422, opened 2015-10-25
Adds option to track uncovered source files.

### LINKED ISSUE #413: Apply work done on handling uncovered source files from #252
ref: #252

The work there is pretty old and may require some rework. This PR is for you :)

hugopeixoto:
Fixes #413 

Added a configuration option `track_files` that accepts
a glob. With this enabled, coverage result will include
uncovered files that match the glob.

In the `rails` profile this is set to `{app,lib}/**/*.rb`

Regarding line count issue:

It is not easy to determine which lines should be classified as relevant (this is done by the ruby parser), so this implementation does not touch the coverage information for covered files, to avoid ruining the accurate percentages.
When handling uncovered files it uses an approximation function to determine if the line is relevant or not, based on the code present in #16. As uncovered files have 0% coverage, this will only affect the global percentage, so I don't think it's an issue.

### DIFF
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -49,11 +49,44 @@ module SimpleCov
     end
 
     #
+    # Finds files that were to be tracked but were not covered and initializes
+    # their coverage to zero, with an estimation of the line count.
+    #
+    def add_uncovered_files(result)
+      if @track_files_glob
+        result = result.dup
+        Dir[@track_files_glob].each do |file|
+          absolute = File.expand_path(file)
+
+          unless result[absolute]
+            lines = File.readlines(absolute)
+            result[absolute] = [0] * lines.count { |l| relevant_line?(l) }
+          end
+        end
+      end
+
+      result
+    end
+
+    #
+    # Determines if a line should be considered by coverage tools.
+    # This method is just an estimation, but it is not very important
+    # that it is accurate. It is only used when all lines in a file
+    # are not covered. As soon as a single line is covered, the counting
+    # method will be the one provided by the Coverage module.
+    #
+    def relevant_line?(line)
+      l = line.strip
+
+      !(l.empty? || l =~ /^else$/ || l =~ /^end$/ || l[0] == '#' || l =~ /^rescue(\s+.*)?$/)
+    end
+
+    #
     # Returns the result for the current coverage run, merging it across test suites
     # from cache using SimpleCov::ResultMerger if use_merging is activated (default)
     #
     def result
-      @result ||= SimpleCov::Result.new(Coverage.result) if running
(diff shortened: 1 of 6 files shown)


lib/simplecov.rb
```diff
@@ -49,11 +49,44 @@ def start(profile = nil, &block)
     end
 
     #
+    # Finds files that were to be tracked but were not covered and initializes
+    # their coverage to zero, with an estimation of the line count.
```
bf4: `s/covered/loaded`

lib/simplecov.rb
```diff
@@ -49,11 +49,44 @@ def start(profile = nil, &block)
...
+    #
+    def add_uncovered_files(result)
+      if @track_files_glob
```
bf4: I think it's reasonable to say that if you are telling SimpleCov you expect a file to be used, but it isn't even loaded, to count that as zero.

However, I understand your concern that any Ruby file that is loaded has a non zero 'coverage', so perhaps, if the file isn't loaded at the time to results are calculated, we just load the file and generate the coverage for those, as usual?

bf4: perhaps with a flag that it wasn't actually run and its own group?

hugopeixoto: I thought about loading the files, but I was a bit afraid of potential side effects (such as syntax errors, or weird metaprogramming stuff). Though that would be a point in favor of considering each line as not covered, as the file hasn't even been syntax checked. I'm OK with removing those checks and mark every line as not covered. Or would you prefer to load the files?

Personally I would rather have them in their original groups, otherwise I would probably tend to ignore them.

xaviershay:
Looks like this is just about ready - a typo remains. I'm not sure whether the discussion about `@track_files_glob` is resolved?

hugopeixoto:
I'll apply the suggestion to mark the whole file as uncovered, instead of having a function to validate the lines (and fix the typo).

hugopeixoto:
The builds are failing due to unrelated issues which are fixed in https://github.com/colszowka/simplecov/pull/425, by @xaviershay.

xaviershay:
Added some method documentation and merged. Thank you! I'll include this in 0.11.

andyw8:
This seems a really useful feature for finding unused code – worth highlighting in the README?

ClaudioCarmeli:
Hi @andyw8 
> This seems a really useful feature for finding unused code
How so?

PragTob:
@ClaudioCarmeli I think he meant because it finds even files that aren't even required --> those can probably be deleted or something in the setup is wrong/incomplete

andyw8:
Yes, that's what I was thinking of.

PR #423, opened 2015-11-04
Recurse upward looking for .simplecov

alexdowad:
The SimpleCov documentation states: "To avoid [duplicated configuration], you
can place a file called `.simplecov` in your project root." However, SimpleCov did
not actually look in the "project root" for this file, unless specially configured
to know where the project root is. Rather, it would simply look in the working
directory. Working directory != project root. (Unfortunately.)

Therefore, when looking for `.simplecov`, recurse upwards from the working directory
until it is found (or the root directory is reached). This is comparable to the
way that git recurses upwards looking for a .git directory, meaning that you can
use git anywhere in your project root or below.

Since there is no guarantee that you have permissions to read `.simplecov`, catch
exceptions and display a warning message if one occurs.

### DIFF
--- a/lib/simplecov/defaults.rb
+++ b/lib/simplecov/defaults.rb
@@ -1,5 +1,6 @@
 # Load default formatter gem
 require "simplecov-html"
+require "pathname"
 
 SimpleCov.profiles.define "root_filter" do
   # Exclude all files outside of simplecov root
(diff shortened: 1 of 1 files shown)


xaviershay:
Prior art:

* RSpec does not traverse up the directory tree. 
* Git does.

This behaviour seems ok, but I think I'd want RSpec to be consistent (since they're so often used together).

alexdowad:
@xaviershay Note that the testing process for a Ruby project is very unlikely to involve running RSpec multiple times, from different subdirectories. `git`, on the other hand, is very frequently used from a subdirectory.

This patch was prompted by work being done on [Racc](https://github.com/tenderlove/racc). Its test suite runs the `./bin/racc` script many times, with various inputs, and checks the output. Naturally, `./bin/racc` runs with a different working directory than the Rake process which invokes it.

I think that this usecase is not entirely unique to Racc. Many test suites may run Ruby subprocesses, for one reason or another. Those subprocesses may or may not have the same working directory.

alexdowad:
Fixed up the style errors.

xaviershay:
Last chance for comment, I'm going to merge this later today otherwise.

Issue #428, opened 2015-12-01
0.11.0 deprecates SimpleCov::Formatter::MultiFormatter.new#[] in a breaking way.


envygeeks:
0.11.0 claims to merely deprecate `SimpleCov::Formatter::MultiFormatter#[]` but it actually breaks it in the process of doing so, as the behavior used to be `SimpleCov::Formatter::MultiFormatter.new[Formatter, Formatter]` but it takes that and it expands that array into a set of arguments, instead of passing those arguments as an array like `SimpleCov::Formatter::MultiFormatter.new` expects it to be, unless there is a much bigger semantic problem here.  This change already broke dozens of our projects without notice and it could end up breaking hundreds of others.

dillondrobena:
I noticed this problem as well. I think the issue is line `20` of `simplecov/lib/simplecov/formatter/multi_formatter.rb` which is
 `@formatters ||= Array(formatters)`
 which isn't correct syntax for a new array, this should be changed to something like 
`@formatters ||= Array([formatters])`
Should be a simple fix

xaviershay:
gargh, sorry I didn't notice this in the diff. I'll get a teeny change out asap.

xaviershay:
Fixed in 0.11.1

PR #431, opened 2015-12-02
#428 Fixed issue with arguments

dillondrobena:
The `self.[]` method was passing in multiple arguments to `self.new`, which is only expecting 1. Simply changing it to pass in array seems to work.


xaviershay:
Want to try putting a spec around it?

xaviershay:
never mind, I got it. Want to get a release out tonight.

xaviershay:
rebased and added spec in https://github.com/colszowka/simplecov/pull/432

thank you!

Issue #433, opened 2015-12-02
Comment lines in unloaded files are marked as uncovered


kmewhort:
In version 0.10, the additional lines on mult-iline statements, as well as comments, were ignored -- which seems like the correct behaviour.  In version 0.11, these lines are marked as not covered.

Before and after:
![simplecov_0 10](https://cloud.githubusercontent.com/assets/402654/11548041/7e7fdefe-990d-11e5-937b-c95c118669b3.png)
![simplecov_0 11](https://cloud.githubusercontent.com/assets/402654/11548044/82820f9a-990d-11e5-8008-b7515092e58c.png)

xaviershay:
yeah that's weird, I wouldn't have expected that to change. I'll have to go back through commits to figure it out.

bf4:
@kmewhort Would you like to submit a failing test for us?  It should be pretty easy to identify it with a simple git bisect.

Also, which Ruby are you on, and which profile/config are you running?

(Does changing the method invocation `.` to be trailing make a difference?)

kmewhort:
I was having some trouble reproducing this in a simplified failure-case spec -- and it turns out the issue is a bit more subtle than the one I first stated.  The actual issue is that files with 0% coverage report *ALL* lines as missed, not just the *relevant* lines.

In my particular case, my specs hit 100% coverage for `file_a.rb`, and simplecov correctly reports the relevant lines as hit (reporting nothing for the comment lines and multi-lines).  Then, my cucumber test suite hits 0% coverage of `file_a.rb`, and -- problematically -- simplecov marks ALL the lines as missed, including the comment lines and multilines.  The results I showed in the images above were actually the result of an lcov merge between these two test suites.

Looks like the likely culprit is 81e2431d4047f921309f5f3e6abceb89692c0e95.

xaviershay:
Thanks @kmewhort , the "double run" scenario sounds like enough for us to track it down. @hugopeixoto any thoughts on how we might address this?

hugopeixoto:
If someone needs a quick workaround, setting `track_files nil` in their simplecov configuration code disables the unloaded files feature.

It seems that it is the merging process that's causing the problem. The coverage from each run probably look something like this:

    spec: [nil, 1, 1, 1, nil, nil, nil, nil, nil, 1, nil]
    cucumber: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]

When merging these two coverage arrays, the nils have the lowest precedence, so the final coverage is

[...]

PR #436, opened 2015-12-05
Fix unloaded files merge

hugopeixoto:
A test case and an attempt at fixing #433

### DIFF
--- a/spec/merge_helpers_spec.rb
+++ b/spec/merge_helpers_spec.rb
@@ -25,7 +25,7 @@ describe "merge helpers" do
       end
 
       it "has proper results for sample.rb" do
-        expect(subject[source_fixture("sample.rb")]).to eq([1, 1, 2, 2, nil, nil, 2, 2, nil, nil])
+        expect(subject[source_fixture("sample.rb")]).to eq([nil, nil, 2, 2, nil, nil, 2, 2, nil, nil])
       end
 
       it "has proper results for user.rb" do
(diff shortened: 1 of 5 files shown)


xaviershay:
I'm not familiar enough to review this properly - the model seems weird but this change seems reasonable without a more extensive refactoring (if that's even appropriate).

Will wait for either someone else to chime in or a confirmation that this has fixed #433 before merging.

Thank you!

spec/merge_helpers_spec.rb
```diff
@@ -25,7 +25,7 @@
       end
 
       it "has proper results for sample.rb" do
-        expect(subject[source_fixture("sample.rb")]).to eq([1, 1, 2, 2, nil, nil, 2, 2, nil, nil])
+        expect(subject[source_fixture("sample.rb")]).to eq([nil, nil, 2, 2, nil, nil, 2, 2, nil, nil])
```
sinsoku: Why did you change the existing behavior?
This will occur a problem when we run the tests in parallel. Because covered line become a disabled line.

bfontaine:
Just FYI there’s a different fix for the same issue:
```patch
diff --git i/lib/simplecov/merge_helpers.rb w/lib/simplecov/merge_helpers.rb
index 339210d1..ca6b8a9f 100644
--- i/lib/simplecov/merge_helpers.rb
+++ w/lib/simplecov/merge_helpers.rb
@@ -2,6 +2,12 @@ module SimpleCov
   module ArrayMergeHelper
     # Merges an array of coverage results with self
     def merge_resultset(array)
+      # Don't merge with all-zero array to preserve nil values from actual runs.
+      if size == array.size
+        return array.dup if all? { |el| el == 0 }
+        return dup if array.all? { |el| el == 0 }
+      end
+
       new_array = dup
       array.each_with_index do |element, i|
         if element.nil? && new_array[i].nil?
```

See https://github.com/Homebrew/homebrew/pull/48250#issuecomment-173171990.

xaviershay:
Closing in favour of #441

PR #437, opened 2015-12-07
Fix root_filter evaluates SimpleCov.root before initialization.

tmtm:
If SimpleCov.root is set to other directory, but root_filter uses current directory.

```ruby
require 'simplecov'
SimpleCov.root '/other/directory'
obj = Struct.new(:filename).new('/other/directory/test.rb')
p SimpleCov.filters.any?{ |filter| filter.matches? obj }
#=> returns true but it should be false.
```

### DIFF
--- a/lib/simplecov/defaults.rb
+++ b/lib/simplecov/defaults.rb
@@ -4,8 +4,9 @@ require "pathname"
 
 SimpleCov.profiles.define "root_filter" do
   # Exclude all files outside of simplecov root
-  root_filter = /\A#{Regexp.escape(SimpleCov.root)}/io
+  root_filter = nil
   add_filter do |src|
+    root_filter ||= /\A#{Regexp.escape(SimpleCov.root)}/io
     !(src.filename =~ root_filter)
   end
 end


lib/simplecov/defaults.rb
```diff
@@ -4,8 +4,9 @@
 
 SimpleCov.profiles.define "root_filter" do
   # Exclude all files outside of simplecov root
-  root_filter = /\A#{Regexp.escape(SimpleCov.root)}/io
+  root_filter = nil
   add_filter do |src|
+    root_filter ||= /\A#{Regexp.escape(SimpleCov.root)}/io
     !(src.filename =~ root_filter)
```
bfontaine: It’s unrelated to this PR, but this could be changed into `src.filename !~ root_filter`.

mikemcquaid:
CC @colszowka for thoughts as we're using a patched version of simplecov in Homebrew to workaround this issue. Thanks for your great work on this gem!

UniqMartin:
Just to add a bit more context: This fixes a regression in 0.11+ introduced by #396.

colszowka:
Makes sense, thanks!

lib/simplecov/defaults.rb
colszowka: You're right - if you like please send in a PR :heart:

colszowka:
@mikemcquaid Thank you. We've been collecting a bunch of enhancements and now this bugfix, I'll try to get v0.11.2 out over the weekend - I want to go over the other open PRs and check if we can get those in too. Please complain if it's not out by Monday.

mikemcquaid:
@colszowka Haha, thanks! The other one we (Homebrew) care about is https://github.com/colszowka/simplecov/pull/436; it'd be great if that (or the patch @bfontaine suggested) could make it in (although I know there's merge conflicts on the PR itself...)

bfontaine:
> Please complain if it's not out by Monday.

It’s Monday and it’s not out :smile:

colszowka:
Yip, sorry about that, didn't find the time to properly look into that other pending PR over the weekend since it apparently needs properly reading the source (there is a seemingly legit complaint in there regarding changed behavior)


Either I get that figured out or I'll ship an intermediate 0.11.2 first, then get that sorted and let 0.11.3 follow closely





On Mon, Feb 1, 2016 at 10:56 AM -0800, "Baptiste Fontaine" <notifications@github.com> wrote:













Please complain if it's not out by Monday.




It’s Monday and it’s not out

bfontaine:
@colszowka Don’t worry about the time it takes :smiley:

colszowka:
So, 0.11.2 is out now, I'll try to get the merging thing figured out and let a 0.11.3 follow as soon as possible!

bfontaine:
Thank you!

PR #441, opened 2015-12-16
Fix coverage rate of the parallel_tests

sinsoku:
The coverage rate has decreased when used with `parallel_tests`. So, I fixed it.

## merge_resultset

This Pull Request is to change behaviors of 2 and 4.

| no | element | new_array[i] | before | after |
|----|----|----|----|----|
| 1 | nil | nil | nil | nil |
| 2 | nil | 0   | 0   | **nil** |
| 3 | nil | 1   | 1   | 1 |
| 4 | 0   | nil | 0   | **nil** |
| 5 | 0   | 0   | 0   | 0 |
| 6 | 0   | 1   | 1   | 1 |
| 7 | 1   | nil | 1   | 1 |
| 8 | 1   | 0   | 1   | 1 |
| 9 | 1   | 1   | 2   | 2 |

## Environment

- Ruby 2.2.3
- Rails 4.2.5
- simplecov 0.11.1
- parallel_tests 1.7.0

### DIFF
--- a/lib/simplecov/merge_helpers.rb
+++ b/lib/simplecov/merge_helpers.rb
@@ -5,11 +5,11 @@ module SimpleCov
       new_array = dup
       array.each_with_index do |element, i|
         pair = [element, new_array[i]]
-        if pair.any?(&:nil?) && pair.map(&:to_i).all?(&:zero?)
-          new_array[i] = nil
-        else
-          new_array[i] = element.to_i + new_array[i].to_i
-        end
+        new_array[i] = if pair.any?(&:nil?) && pair.map(&:to_i).all?(&:zero?)
+                         nil
+                       else
+                         element.to_i + new_array[i].to_i
+                       end
       end
       new_array
     end
--- a/lib/simplecov/merge_helpers.rb
+++ b/lib/simplecov/merge_helpers.rb
@@ -4,14 +4,11 @@ module SimpleCov
     def merge_resultset(array)
       new_array = dup
       array.each_with_index do |element, i|
-        if element.nil? && new_array[i].nil?
-          new_array[i] = nil
-        elsif (element.nil? && new_array[i] == 0) || (element == 0 && new_array[i].nil?)
+        pair = [element, new_array[i]]
+        if pair.any?(&:nil?) && pair.map(&:to_i).all?(&:zero?)
           new_array[i] = nil
         else
-          local_value = element || 0
-          other_value = new_array[i] || 0
-          new_array[i] = local_value + other_value
+          new_array[i] = element.to_i + new_array[i].to_i
         end
       end
       new_array
--- a/lib/simplecov/merge_helpers.rb
+++ b/lib/simplecov/merge_helpers.rb
@@ -6,6 +6,8 @@ module SimpleCov
       array.each_with_index do |element, i|
         if element.nil? && new_array[i].nil?
           new_array[i] = nil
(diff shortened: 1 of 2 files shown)


xaviershay:
To be clear: you intend for this to be just adding specs + refactoring, not changing any behaviour?

sinsoku:
No, this Pull Request is changing the behavior a little bit.
Please check the description of Pull Request, because I updated it.

The specification of Coverage module are as below.

- nil => disabled
- 0 => uncovered
- 1 => covered

And the coverage of unloaded files are `[0]` on simplecov v0.11.0.
This is a problem when run with `parallel_tests`, to merges coverages generated by a process.

This Pull Request is intended to solve it.

See also:
https://github.com/colszowka/simplecov/blob/v0.11.0/lib/simplecov.rb#L61
https://github.com/ruby/ruby/blob/v2_2_3/ext/coverage/coverage.c

xaviershay:
ah, so I suspect this is a "dup" of https://github.com/colszowka/simplecov/pull/436 ?

I don't have a preference as to which is a better solution - only reason I haven't merged the other one is that I was waiting for someone to confirm it actually fixed the issue :)

sinsoku:
#436 is insufficient to fix this problem. (https://github.com/colszowka/simplecov/pull/436#discussion_r48090042)
Moreover, #436 is including extensive refactoring, so I could not create a patch for it.

ujh:
It would be great if this could get merged. When I run simplecov as part of a normal rspec run I get around 80% test coverage, but for a while I'm down do around 48% when running as part of parallel_tests. With this branch I'm now back to 80%.

xaviershay:
That's the confirmation I was looking for :)

ujh:
That's why I posted it. 😁

rdunlop:
I just used this change on my project, and it fixed our issue of multiple runs being merged incorrectly, leading to lower-than-actual results.  It would be really nice to get this released in a new version of the gem.

sbleon:
Just added Simplecov to a Rails 4.2 project. I was disturbed to see a much-lower-than-expected coverage percentage, then I figure out that all of my lines of comments were being reported as "missed". Using this branch fixes the problem. Please merge and release it!

lib/simplecov/merge_helpers.rb
```diff
@@ -4,13 +4,12 @@ module ArrayMergeHelper
...
       array.each_with_index do |element, i|
-        if element.nil? && new_array[i].nil?
-          new_array[i] = nil
-        else
-          local_value = element || 0
-          other_value = new_array[i] || 0
-          new_array[i] = local_value + other_value
-        end
+        pair = [element, new_array[i]]
+        new_array[i] = if pair.any?(&:nil?) && pair.map(&:to_i).all?(&:zero?)
```
bf4: so, the result is `nil` if either set contains nil AND both are each one of either nil, zero, or some non-integer?

bf4:
@sinsoku can you rebase and force push your commits, or @colszowka wanna change the merge options? :)

sinsoku:
@bf4 Thank you for the feedback. I rebased and force pushed.

sbleon:
I just wanted to mention that I have the "incorrectly low coverage" problem but I am not using `parallel_tests`. Again, this branch fixes the problem for me, so I'd love to see it merged and released soon. Thanks!

bf4:
@sinsoku @sbleon merged

sbleon:
Thanks!

grosser:
release ? :)

PR #442, opened 2015-12-16
add a group for Active Jobs classes to default Rails profile

stve:
I'm using Active Job in a Rails project for the first time and my job classes are currently "Ungrouped". Since `app/jobs` is the default location for these classes, I think it's worth adding this to the Rails profile.

Are there tests for this? I couldn't find any but if I've missed them, please point them out and I'll add a test as well. Thanks

### DIFF
--- a/lib/simplecov/defaults.rb
+++ b/lib/simplecov/defaults.rb
@@ -31,6 +31,7 @@ SimpleCov.profiles.define "rails" do
   add_group "Models", "app/models"
   add_group "Mailers", "app/mailers"
   add_group "Helpers", "app/helpers"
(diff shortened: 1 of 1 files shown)


colszowka:
Thanks, makes sense to have this!

Since with the existing frameworks like sidekiq and resque, the convention is to go with "app/workers", could you please add this dir as well? You can pass in an array of paths, so `add_group "Jobs", %w(app/jobs app/workers)` should do it.

stve:
Hi @colszowka, thanks for the feedback. I updated the pull request based on your suggestion and rebased against master so that it will merge cleanly.

colszowka:
Thank you @stve !

PR #447, opened 2016-01-20
Silence `@track_files_glob` ivar warning

### LINKED ISSUE #446: lib/simplecov.rb:60: warning: instance variable @track_files_glob not initialized
Hello,

It seems Simplecov module expect some instance variables to be set up by SimpleCov::Configuration. Of course the default value nil works, however, we get warnings from Ruby VM.  I believe this could be solved by improving SimpleCov::Configuration to either preset unset variables to nil, or, check them with Object#instance_variable_defined? before using.

If you think any of these is a good solution (or even suggest a better one), I can make the patch an pull request.

Sincerely,

craiglittle:
Fixes #446.

### DIFF
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -57,9 +57,9 @@ module SimpleCov
     # their coverage to zero.
     #
     def add_not_loaded_files(result)
-      if @track_files_glob
+      if track_files
         result = result.dup
-        Dir[@track_files_glob].each do |file|
+        Dir[track_files].each do |file|
           absolute = File.expand_path(file)
 
           result[absolute] ||= [0] * File.foreach(absolute).count
--- a/lib/simplecov/configuration.rb
+++ b/lib/simplecov/configuration.rb
@@ -47,8 +47,9 @@ module SimpleCov
     # or not they were explicitly required. Without this, un-required files
     # will not be present in the final report.
     #
-    def track_files(glob)
-      @track_files_glob = glob
+    def track_files(glob = nil)
+      return @track_files if defined?(@track_files) && glob.nil?
+      @track_files = glob
     end
 
     #


colszowka:
The build failed due to #448 apparently, we'll have to get that one out of the door first.

We've been dealing with those pesky ruby warnings basically since the beginning of the project, thanks for taking the time to take this one out!

In order to keep things consistent, could you please change the implementation to what we're doing for all the other instance vars in the configuration class? See for example `https://github.com/craiglittle/simplecov/blob/fix-ivar-warning/lib/simplecov/configuration.rb#L36-L37`, basically we return if the var is defined and truthy, otherwise do the assignment. I would prefer not to introduce another pattern for dealing with these things here. It's arguable though whether your solution isn't nicer/cleaner - if you prefer I'd also be good if we changed the existing methods to use your style of pre-initializing on extend.

craiglittle:
Sure, I'll change the other methods to pre-initialize the instance variables in the `extended` hook. FWIW, I leaned this direction because of [@xaviershay's comment in the related issue](https://github.com/colszowka/simplecov/issues/446#issuecomment-173047701).

craiglittle:
Okay, I decided to go the other direction and switch the `track_files_glob` implementation to mirror the other methods because doing otherwise would change the semantics of the other methods.

Once we get the RuboCop PR squared away, I'll rebase this onto latest master.

colszowka:
Perfect, thank you!

xaviershay:
apologies for the bad intel, I didn't realise we had prior art!

colszowka:
It's not the best style, but this build failed solely because of Rubycop, which was fixed in #451, so I'm merging this to avoid having to make another roundtrip with @craiglittle for rebasing this against master

craiglittle:
Do you think a release will be cut soon? It'd be swell to get this warning out of my build. :)

colszowka:
@craiglittle Yes, it's out in 0.11.2 right now

craiglittle:
Awesome, much appreciated! 😄

lib/simplecov.rb
```diff
@@ -57,9 +57,9 @@ def start(profile = nil, &block)
     # their coverage to zero.
     #
     def add_not_loaded_files(result)
-      if @track_files_glob
+      if track_files
```
kmewhort: I think this needs to be `@track_files` (as I don't see an accessor method for track_files defined anywhere).

We have a `track_files nil` configuration in our .simplecov that seems to be skipped after upgrading to the latest version, which led me here.

lib/simplecov.rb
```diff
@@ -57,9 +57,9 @@ def start(profile = nil, &block)
...
         result = result.dup
-        Dir[@track_files_glob].each do |file|
+        Dir[track_files].each do |file|
```
kmewhort: Same here.

PR #448, opened 2016-01-20
Make compatible with latest RuboCop

craiglittle:
This brings it up-to-date for `v0.36.0`.

Changelog: https://github.com/bbatsov/rubocop/blob/master/CHANGELOG.md#0360-14012016

### DIFF
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -74,14 +74,22 @@ module SimpleCov
     # from cache using SimpleCov::ResultMerger if use_merging is activated (default)
     #
     def result
-      @result ||= SimpleCov::Result.new(add_not_loaded_files(Coverage.result)) if running
+      # Ensure the variable is defined to avoid ruby warnings
+      @result = nil unless defined?(@result)
+
+      # Collect our coverage result
+      if running && !result?
+        @result = SimpleCov::Result.new add_not_loaded_files(Coverage.result)
+      end
+
       # If we're using merging of results, store the current result
       # first, then merge the results and return those
       if use_merging
-        SimpleCov::ResultMerger.store_result(@result) if @result
-        return SimpleCov::ResultMerger.merged_result
+        SimpleCov::ResultMerger.store_result(@result) if result?
+
+        SimpleCov::ResultMerger.merged_result
       else
-        return @result if defined? @result
+        @result
       end
     ensure
       self.running = false
--- a/lib/simplecov/version.rb
+++ b/lib/simplecov/version.rb
@@ -1,22 +1,25 @@
 module SimpleCov
-  VERSION = "0.11.1"
-  def VERSION.to_a
+  version = "0.11.1"
+
+  def version.to_a
     split(".").map(&:to_i)
   end
 
-  def VERSION.major
+  def version.major
     to_a[0]
   end
 
-  def VERSION.minor
+  def version.minor
     to_a[1]
   end
 
-  def VERSION.patch
+  def version.patch
     to_a[2]
   end
 
-  def VERSION.pre
+  def version.pre
     to_a[3]
   end
+
+  VERSION = version.freeze
 end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -68,3 +68,6 @@ Style/FileName:
 
 Style/TrailingCommaInLiteral:
   EnforcedStyleForMultiline: comma
+
+Style/GuardClause:
+  Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -66,5 +66,5 @@ Style/FileName:
   Exclude:
     - 'spec/fixtures/utf-8.rb'
 
-Style/TrailingComma:
-  EnforcedStyleForMultiline: 'comma'
+Style/TrailingCommaInLiteral:
+  EnforcedStyleForMultiline: comma
--- a/Gemfile
+++ b/Gemfile
@@ -26,7 +26,7 @@ group :test do
     gem "cucumber", "~> 2.0"
     gem "phantomjs", "~> 1.9"
     gem "poltergeist", "~> 1.1"
-    gem "rubocop", ">= 0.30"
+    gem "rubocop", "~> 0.36.0"
     gem "test-unit", "~> 3.0"
   end
 end
--- a/lib/simplecov/configuration.rb
+++ b/lib/simplecov/configuration.rb
@@ -114,7 +114,7 @@ module SimpleCov
       return @nocov_token if defined?(@nocov_token) && nocov_token.nil?
       @nocov_token = (nocov_token || "nocov")
     end
-    alias_method :skip_token, :nocov_token
+    alias skip_token nocov_token
 
     #
     # Returns the configured groups. Add groups using SimpleCov.add_group
--- a/lib/simplecov/defaults.rb
+++ b/lib/simplecov/defaults.rb
@@ -51,15 +51,16 @@ at_exit do
   # If we are in a different process than called start, don't interfere.
   next if SimpleCov.pid != Process.pid
 
-  if $! # was an exception thrown?
-    # if it was a SystemExit, use the accompanying status
-    # otherwise set a non-zero status representing termination by some other exception
-    # (see github issue 41)
-    @exit_status = $!.is_a?(SystemExit) ? $!.status : SimpleCov::ExitCodes::EXCEPTION
-  else
-    # Store the exit status of the test run since it goes away after calling the at_exit proc...
-    @exit_status = SimpleCov::ExitCodes::SUCCESS
-  end
+  @exit_status = if $! # was an exception thrown?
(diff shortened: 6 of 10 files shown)


lib/simplecov.rb
```diff
@@ -79,9 +79,9 @@ def result
       # first, then merge the results and return those
       if use_merging
         SimpleCov::ResultMerger.store_result(@result) if @result
-        return SimpleCov::ResultMerger.merged_result
-      else
-        return @result if defined? @result
+        SimpleCov::ResultMerger.merged_result
+      elsif defined? @result
+        @result if defined? @result
```
colszowka: I like the approach, but this changes the logic, you're missing an `else SimpleCov::ResultMerger.merged_result` case here to keep it in line with the previous version, no?

lib/simplecov/version.rb
```diff
@@ -1,22 +1,25 @@
 module SimpleCov
-  VERSION = "0.11.1"
-  def VERSION.to_a
+  version = "0.11.1"
```
colszowka: Why not directly VERSION = '123'.freeze?

lib/simplecov.rb
craiglittle: Whoops, I think the right move is to get rid of the now-redundant `if defined? @result`. On the way.

lib/simplecov/version.rb
colszowka: Ah, I see, you're adding methods to the object afterwards. Do we need this?

spec/faked_project/lib/faked_project/some_class.rb
```diff
@@ -11,11 +11,9 @@ def reverse
   end
 
   def compare_with(item)
-    if item == label
-      return true
-    else
-      fail "Item does not match label"
-    end
+    return true if item == label
```
colszowka: I know this is not applied to all of the code base and this especially is "just" our test sample code, but personally I prefer the actual if/else form because it's easier to parse the control flow of the method

lib/simplecov/version.rb
craiglittle: There's a new cop that requires all literals assigned to constants to be frozen. Because the method definitions that follow modify the `VERSION` string, I had to do these gymnastics to make everything happy.

Perhaps we should create a `Version` module and make those methods class methods of that module?

I'm cool with going with whatever approach you all prefer!

spec/faked_project/lib/faked_project/some_class.rb
craiglittle: Sure. Would you like me to disable that cop completely or special-case this file?

lib/simplecov/version.rb
craiglittle: Or, I can disable that cop.

colszowka: Ah, dammit - I had overlooked that we already were modifying the const object with those methods. You're right, let's keep it this way!

spec/faked_project/lib/faked_project/some_class.rb
colszowka: Which one is it and what is it complaining about? Can't see anything in the build log :(

lib/simplecov.rb
colszowka: Reading this method makes me dizzy, maybe we get to untangle this a bit:

```ruby
def result
  # Ensure the variable is defined to avoid ruby warnings
  @result = nil unless defined?(@result)

  # Collect our coverage result
  if running and not result?
    @result = SimpleCov::Result.new add_not_loaded_files(Coverage.result)
  end

  # If we're using merging of results, store the current result
  # first, then merge the results and return those
  if use_merging
    if result?
      SimpleCov::ResultMerger.store_result @result
    end
    SimpleCov::ResultMerger.merged_result
  else
    @result
  end
ensure
  self.running = false
end
```

spec/faked_project/lib/faked_project/some_class.rb
craiglittle: That would be the [`Style::GuardClause` cop](http://www.rubydoc.info/github/bbatsov/rubocop/Rubocop/Cop/Style/GuardClause), which says:
```
Use a guard clause instead of wrapping the code inside a conditional expression.
```

craiglittle:
I've added the refactor of the `result` method with a few tweaks to continue to make RuboCop happy. Now, all we need to resolve is what to do with the guard clause.

craiglittle:
Also, I think it'd be prudent to consider locking `rubocop` to a minor version so it doesn't randomly break the build whenever a new version is released. Then, it could be updated when it's convenient instead of blocking other PRs.

Would you be open to me making that change?

craiglittle:
@colszowka When you get a chance, I'm waiting on feedback on two points in order to proceed:
* What should we do about the [`Style/GuardClause` cop](https://github.com/colszowka/simplecov/pull/448#discussion_r50287461)?
* Should we [lock down `rubocop`](https://github.com/colszowka/simplecov/pull/448#issuecomment-173314414) to avoid unexpected build failures in the future?

colszowka:
Thanks for keeping up on this thing!


* As I mentioned before, nowadays I personally prefer explicit conditionals over guard clauses because imho they're easier to read. But that also requires you to make your conditionals clean. I think in this case that is true, but its my personal opinion. Maybe you could create a follow-up issue where we bring in the rest of the contributors team to figure this out, and disable it for now to resolve the PR?

* yes please, I was thinking about this myself. Please lock it at patch level since we're currently at 0.36 if I remember correctly, so ~>0.36 wouldnt fix the problem :)


Thanks again





On Fri, Jan 22, 2016 at 2:14 PM -0800, "Craig Little" <notifications@github.com> wrote:












@colszowka When you get a chance, I'm waiting on feedback on two points in order to proceed:


What should we do about the Style/GuardClause cop?
Should we lock down rubocop to avoid unexpected build failures in the future?

craiglittle:
> As I mentioned before, nowadays I personally prefer explicit conditionals over guard clauses because imho they're easier to read. But that also requires you to make your conditionals clean. I think in this case that is true, but its my personal opinion. Maybe you could create a follow-up issue where we bring in the rest of the contributors team to figure this out, and disable it for now to resolve the PR?

[Done](https://github.com/colszowka/simplecov/issues/452) and [done](https://github.com/colszowka/simplecov/commit/dfe701bea243f6b81a3c0e42a645bc2ef18f47f4)!

> yes please, I was thinking about this myself. Please lock it at patch level since we're currently at 0.36 if I remember correctly, so ~>0.36 wouldnt fix the problem :)

[Locked](https://github.com/craiglittle/simplecov/commit/c4bc69f3a60952fe57be2aa023a4982ed5bb7c29) to the `v0.36.x` series.

And the build's :green_heart:!

colszowka:
Perfect, thanks @craiglittle

lib/simplecov/version.rb
leafle: I think freezing the version constant is actually breaking the simplecov-teamcity-summary with older versions of rubygems. The simplecov-teamcity-summary has the following code:
`Gem::Version.new(SimpleCov::VERSION)`
and rubygems in version 1.8.23 has `@version.strip!` in its constructor. In newer versions of rubygems they're using `@version.strip` so I don't think it's an issue there.

leafle: The following is also broken with that rubygems version:

```
$ git clone https://github.com/colszowka/simplecov.git
...
$ cd simplecov/
$ bundle
[!] There was an error parsing `Gemfile`: There was a RuntimeError while loading simplecov.gemspec: 
can't modify frozen String from
  ~/Workspace/simplecov/simplecov.gemspec:6:in `block in <main>'
. Bundler cannot continue.

 #  from ~/Workspace/simplecov/Gemfile:34
 #  -------------------------------------------
 #  
 >  gemspec
 #  -------------------------------------------
$ gem --version
1.8.23.2
$ ruby --version
ruby 1.9.3p551 (2014-11-13 revision 48407) [x86_64-darwin14.5.0]
```

colszowka: Oh wow, there are the most amazing bugs from the simplest things. 

I'm somewhat torn on this one because (see https://github.com/colszowka/simplecov/issues/458) we basically shouldn't be spending time on supporting rubies that were [EOLd a year ago](https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/) but this is such a relatively small thing to fix and make things work on old rubies again. Maybe simplecov should even outright refuse to be compatible with outdated rubies to make people aware of the fact they're exposing themselves to unfixed security and performance problems.

Anyway, could you please send in a PR that disables the relevant Cop, with a comment about why it's disabled and a link to here as well as the unfreezing of the version string? Also, please upgrade your Ruby, in my experience the upgrades from 1.9 on are very easy.

colszowka: Fixed in https://github.com/colszowka/simplecov/pull/461

PR #449, opened 2016-01-20
Stop polluting ::Array and ::Hash

amatsuda:
Requiring simlpecov globally adds the `merge_resultset` method to the built-in Array class and Hash class via monkey-patching, which means this gem widely changes the test subject's behavior.
Such a horrible vandalism should never be done by a test coverage measurement tool. We should immediately stop doing this.

I suppose the same merging logic could be more safely and cleanly implemented by two or three static methods instead of extending the core classes, but for now, I'd suggest a patch that changes the Result class to dynamically extend the result objects using the existing Modules.

This way we would be able to minimize the conflict with other PRs working on the merging_helpers e.g. #436 and #441.

### DIFF
--- a/lib/simplecov/merge_helpers.rb
+++ b/lib/simplecov/merge_helpers.rb
@@ -27,12 +27,9 @@ module SimpleCov
       end
 
       new_resultset.each_key do |filename|
-        new_resultset[filename] = (self[filename] || []).merge_resultset(hash[filename] || [])
+        new_resultset[filename] = (self[filename] || []).extend(SimpleCov::ArrayMergeHelper).merge_resultset(hash[filename] || [])
       end
       new_resultset
     end
   end
 end
-
-Array.send :include, SimpleCov::ArrayMergeHelper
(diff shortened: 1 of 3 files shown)


colszowka:
The build failed because of Rubycop, there is #448 for that.

colszowka:
As to the actual topic: I totally agree with you, we must stop doing this, thanks for pointing it out!

We still need to remove the actual includes from the source https://github.com/colszowka/simplecov/blob/master/lib/simplecov/merge_helpers.rb#L37-L38 - would you mind doing this? I would expect that with the existing changes this should keep working?

amatsuda:
@colszowka 
> We still need to remove the actual includes from the source

I believe that was actually included in the previous commit, and at least the specs and features were passing locally with that changeset!

colszowka:
I'm an incapable of reading diffs it seems. Thanks! Merging! I also created #450 for the followup work

colszowka:
Lol, awesome branch naming :monkey_face: :gun:

PR #451, opened 2016-01-22
Fixed warnings

fornellas:
Fix for #446

### DIFF
--- a/lib/simplecov/configuration.rb
+++ b/lib/simplecov/configuration.rb
@@ -10,6 +10,12 @@ module SimpleCov
   module Configuration # rubocop:disable ModuleLength
     attr_writer :filters, :groups, :formatter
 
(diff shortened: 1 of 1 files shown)


colszowka:
Thanks for looking into this @fornellas - we do have an existing PR for this fix by now already though with #447, which also uses the same strategy for initializing the var as we have used in the rest of the methods in the module. Only thing it's waiting for is getting #448 ready - sorry for the double work!

fornellas:
No problem, glad to help :-D

fornellas:
BTW, I'd consider enabling warnings in RSpec, to avoid future mistakes.

colszowka:
@fornellas Yes, I thought so too - I think at one point we did have those, but maybe I'm mistaken. Would you mind sending in a PR with this? :heart:

PR #455, opened 2016-02-03
nocov_regex feature

devvmh:
If you add regular expressions like this:

```
SimpleCov.start do
  nocov_regex /^[\s]*annoying_method/
end
```

then lines matching the regular expression will be skipped, just as if
they were inside a :nocov: block.

I'm hoping to be able to use this for a project. I've got a number of controllers that render errors in certain cases. I don't want to write rspec tests for every single possible error case, and I know the render_error function works, so I'm hoping to be able to just filter out lines that call that function (but not the lines of the function itself).

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,6 +333,16 @@ The name of the token can be changed to your liking. [Learn more about the nocov
 
 **Note:** You shouldn't have to use the nocov token to skip private methods that are being included in your coverage. If you appropriately test the public interface of your classes and objects you should automatically get full coverage of your private methods.
 
+You can also exclude code using a regular expression. Setup your .simplecov file like this:
+
+```ruby
+SimpleCov.start do
+  nocov_regex /^[\s]*annoying_method/
+end
+```
+
+and lines matching the regular expression will be skipped, just as if they were inside a :nocov: block. In this example, the code for annoying_method would be tested ("def annoying_method"), but any call to that method would be skipped. This may be useful for code that renders errors, if you don't want to write tests for every single error handling condition.
+
 ## Default root filter and coverage for things outside of it
 
 By default, SimpleCov filters everything outside of the `SimpleCov.root` directory. However, sometimes you may want
--- a/features/skipping_code_blocks_automatically.feature
+++ b/features/skipping_code_blocks_automatically.feature
@@ -1,0 +1,39 @@
+@test_unit @nocov_regex
+Feature:
+
+  When code matches a nocov regex, it does not count against the coverage numbers
+
+  Background:
+    Given SimpleCov for Test/Unit is configured with:
+      """
+      require 'simplecov'
+      SimpleCov.start 'test_frameworks' do
+        SimpleCov.nocov_regex /^[\s]*def some_weird_code/
+        SimpleCov.nocov_regex /^[\s]*never_reached/
+      end
+      """
+
+  Scenario: Plain run with a nocov'd method
+    Given a file named "lib/faked_project/nocov_regex.rb" with:
+      """
+      class SourceCodeMatchingNocovRegex
+        def some_weird_code
+          never_reached
+        end
+      end
+      """
+
+    When I open the coverage report generated with `bundle exec rake test`
+
+    Then I should see the source files:
+      | name                                    | coverage |
+      | lib/faked_project.rb                    | 100.0 %  |
+      | lib/faked_project/some_class.rb         | 80.0 %   |
+      | lib/faked_project/framework_specific.rb | 75.0 %   |
+      | lib/faked_project/meta_magic.rb         | 100.0 %  |
+      | lib/faked_project/nocov_regex.rb        | 100.0 %  |
+
+    And there should be 2 skipped lines in the source files
+
+    And the report should be based upon:
+      | Unit Tests |
--- a/features/skipping_code_blocks_automatically.feature
+++ b/features/skipping_code_blocks_automatically.feature
@@ -1,4 +1,4 @@
-@test_unit @nocov_regex
+@test_unit
 Feature:
 
   When code matches a nocov regex, it does not count against the coverage numbers
--- a/lib/simplecov/configuration.rb
+++ b/lib/simplecov/configuration.rb
@@ -128,8 +128,7 @@ module SimpleCov
     # this method multiple times to define multiple regular expressions
     #
     def nocov_regexes(nocov_regex = nil)
-      return @nocov_regexes if defined?(@nocov_regexes) && nocov_regex.nil?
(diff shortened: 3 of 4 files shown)


xaviershay:
Thanks for the suggestion! It's a neat implementation also.

I have some minor concerns about this:

* Expands configuration interface (you kind of have to do this for a new feature, but still)
* Makes it easier to create "magic" action at a distance. If I encounter a new project, how do I figure out why some method isn't getting code coverage? A `nodoc` comment is right there in front of me. This is happening somewhere else I won't expect.

Wonder what others think?

If we do merge this, we'll need to include something in the README also.

devvmh:
Well, I added a README explanation. Anyone else have thoughts on this PR?

I understand both points you raise and yield to you guys on design decisions :D. I can keep using my fork if I have to, but it is nicer to pull directly from rubygems :smiley:

devvmh:
Another approach would be to configure these regexes in comments at the top of individual files. That would address your second concern about spooky action at a distance. I would actually lean towards both options - global nocov regexes and file-specific ones. If this PR gets merged I could implement the comment part too

PragTob:
Hi there!

Thanks for taking the work to add a feature to simplecov :tada: & sorry for the silence :)

So my first thought on this was "cool, let's figure out a way to do this" but the more I thought about it, the less I liked the feature. I think `:nocov:` and friends should be used sparsely and if they are used it should be obvious, explicit and maybe even a little bit painful :D 

E.g. the use case you mentioned in the README specifically says you don't want to test all those error cases. That's a decision you make, and imo it's fine to be reflected in the coverage. You don't have to have 100% test coverage and can live with less.

I believe skipping too much leads to a false sense of security where you feel secure as you think you covvered 99,x% but in reality it's "just" 88% or something. Hence, I'd like to discourage this behaviour.

All that said I haven't been with the project for long (although longtime simplecov user) so it's certainly not my call, these are just my thoughts on the feature.

devvmh:
Well, I guess I'll have to keep my fork up to date. At my past workplace, we were using simplecov as one way to measure our technical debt and aiming for a fixed threshold. I don't think that's an uncommon approach, although probably better workplaces would be smarter about it. 

This feature allowed me to save significant amounts of developer time, so junior developers didn't need to write test cases for every error handler for every API endpoint of our rails app.

That said, I do appreciate the concept of having less configuration, as put so succintly in https://github.com/jlongster/prettier/issues/40. And as I said above, I do yield to you guys on design decisions for your gem.

I could probably be convinced to rework this to use regular expressions at the top of the file, or to try to create a fourth class of lines called "explicitly skipped" or something to alleviate those concerns. But if it won't be merged anyways, I'm happy for this PR to be closed, and I'll just rebase my fork off of the master branch every now and then.

PragTob:
Hey there once again!

Thanks for engaging in the discussions :)

Could you be more explicit about your use case? E.g. I'm still a bit unsure, from what I'm thinking right now it's something like:

* set minimum coverage to 80%
* PRs (esp. by Juniors) would lead to coverage drops as not all error cases tested --> failing builds --> frustration
* decision that that's not too important for product/team/whatever --> exclude some error cases/calls from coverage as other stuff more important right now?

Does that sort of adequately summarize your problem? 

I don't think another item will help - skipped is already explicitly skipped :)

So still a bit unsure about this, but you said the magic word (junior devs) and working off a fork doesn't seem satisfying + I believe that although I might not directly support the use case, great libraries offer extension points so that you could implement the functionality in your code base/plugin not in a fork. Currently thinking maybe passing in a block/generic object.

Will wait for your reply and then maybe go and ask Christoph and see what he thinks :)

colszowka:
Thanks for your contribution @devvmh !

I agree with the feedback @PragTob and @xaviershay gave that this might be encouraging "wrong" behaviour, in that it should always be clear what is covered or not. Part of the allure of `:nocov:` is that it is very ugly to see, and it should be nagging developers to write some tests. Introducing a less-explicit way will skew your metrics and nobody will ever know why.

To also give some feedback on your example, in my opinion it is exactly the error handling cases which tend to get overseen for tests and at the same time are very critical to test, since when something already went wrong you don't want it to go wrong again by having a typo in an error handler method call or whatnot, especially if it can only go wrong in production (due to missing tests). These cases are usually also pretty simple to test by mocking the actual cause of the exception, i.e. `expect(Widget).to receive(:update_attributes).and_raise(NoMethodError)` and only verifying the exception handling, not the logic that triggers that exception in the first place.

I also don't see a need to have a specific minimum threshold of coverage except for political reasons, do whatever works for your team. You can also use the maximum coverage drop feature instead, by simply disallowing the reduction of coverage on changes. On cloud-based CI servers you will need to cache your coverage results between builds for this to work though.

Speaking about junior devs, I see huge potential in this feature to accidentally change the regexp to match huge swaths of your code - how would you notice your "coverage" of 100% has been a lie for months? With the nocov comment at least it is very clear from reading the code.

Finally, I also agree with Tobi that it should not be neccessary to maintain a fork to accommodate a use case if instead it would be possible to have an extensible API. Tobi suggested to me having an API similar to the filtering/grouping, in that there'd be a default class for NoCov filtering that you can swap out with your custom implementation, i.e.

```
SimpleCov.nocov_filter SimpleCov::NoCov::CommentBlockFilter
```

I don't remember right now what the current implementation of nocov looks like, but it might be a bit messy and due for a cleanup anyway. I also can't really tell off the top of my head what the exact API for nocov classes should look like because of that, but it should be fairly straightforward.

Because of the reasons above I would not ship a Regex filter as part of simplecov though, I'd keep adding this as an intentional extra hoop to jump through after good consideration.

devvmh:
The problem is I have one file with this function:

```
def render_error(args)
  # ... do error rendering in a general way
end
```

This function is all green - I've got tests for it.

Then there are like 50 other files with 5+ actions each. All of them have code looking something like

```
def create
  @record = Record.new(record_params)
  if @record.save
    render json: @record
  else
    render_error                  # <--- this line is red, all others are green.
  end
end
```

Now I have to go write 250 pointless tests. I've got a test for the render error function; I know it works. But there are 250 red lines in my code coverage that are the part where I *call* render_error. Yes, if there are weird use cases I should test this call. But most of my endpoints are really standard, so those tests are just cut and paste tests that waste time in my builds and (assuming I don't script the writing of the tests) waste time for the developers writing them every time we add a new endpoint.

I'm hearing that this won't get merged, and I understand the reasons why and I see the argument. I agree with that argument theoretically, but I still think that in the real world this is a useful & valuable feature. 

But I've moved on from the job where I needed this anyways, so we might as well close this! Thanks everybody for reviewing and adding thoughtful comments.

PR #456, opened 2016-02-03
Tests for warnings

fornellas:
@colszowka , regarding #451, I found a partial solution. It seems Ruby warnings are issued without consistency:
* From Ruby code, using [Kernel#warn](http://ruby-doc.org/core-2.2.3/Kernel.html#method-i-warn)
* From C code, using [rb_warn()](https://github.com/ruby/ruby/blob/439224a5904411b288e441096e21a41244ddd1d6/error.c#L236)
* Printing to $stderr directly [in some cases](https://github.com/ruby/ruby/blob/439224a5904411b288e441096e21a41244ddd1d6/lib/net/ftp.rb#L223)

If warnings are issued within an example block, this PR will detect and fail the test. As  bonus, it will not fail tests if warnings were issued by code at some dependency outside the project.

Warnings issued outside example blocks, for example with a require, will not be detected though. The solution, would be to use the same technique, but inside the Rakefile perhaps.

Anyway, partial solution is better than nothing :-D

### DIFF
--- a/lib/simplecov/configuration.rb
+++ b/lib/simplecov/configuration.rb
@@ -10,12 +10,6 @@ module SimpleCov
   module Configuration # rubocop:disable ModuleLength
     attr_writer :filters, :groups, :formatter
 
-    def self.extended(base)
-      base.instance_eval do
-        @track_files_glob = nil
-      end
-    end
-
     #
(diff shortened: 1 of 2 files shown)


fornellas:
I'll fix Ruby 1.8

PR #457, opened 2016-02-03
Tests for warnings

fornellas:
@colszowka , regarding #451, I found a partial solution. It seems Ruby warnings are issued without consistency:
- From Ruby code, using [Kernel#warn](http://ruby-doc.org/core-2.2.3/Kernel.html#method-i-warn)
- From C code, using [rb_warn()](https://github.com/ruby/ruby/blob/439224a5904411b288e441096e21a41244ddd1d6/error.c#L236)
- Printing to $stderr directly [in some cases](https://github.com/ruby/ruby/blob/439224a5904411b288e441096e21a41244ddd1d6/lib/net/ftp.rb#L223)

If warnings are issued within an example block, this PR will detect and fail the test. As  bonus, it will not fail tests if warnings were issued by code at some dependency outside the project.

Warnings issued outside example blocks, for example with a require, will not be detected though. The solution, would be to use the same technique, but inside the Rakefile perhaps.

Anyway, partial solution is better than nothing :-D

### DIFF
--- a/lib/simplecov/configuration.rb
+++ b/lib/simplecov/configuration.rb
@@ -10,12 +10,6 @@ module SimpleCov
   module Configuration # rubocop:disable ModuleLength
     attr_writer :filters, :groups, :formatter
 
-    def self.extended(base)
-      base.instance_eval do
-        @track_files_glob = nil
-      end
-    end
-
     #
     # The root for the project. This defaults to the
     # current working directory.
--- a/lib/simplecov/configuration.rb
+++ b/lib/simplecov/configuration.rb
@@ -10,6 +10,12 @@ module SimpleCov
   module Configuration # rubocop:disable ModuleLength
     attr_writer :filters, :groups, :formatter
 
+    def self.extended(base)
+      base.instance_eval do
+        @track_files_glob = nil
+      end
+    end
+
     #
     # The root for the project. This defaults to the
     # current working directory.
--- a/spec/helper.rb
+++ b/spec/helper.rb
@@ -22,3 +22,12 @@ ensure
   # Restore the previous value of stderr (typically equal to STDERR).
   $stderr = previous_stderr
 end
+
+RSpec.configure do |config|
+  # Fail tests if code generates runtime warnings
+  config.around(:example) do |example|
+    expect{example.call}
+      .not_to output(/^#{Regexp.escape(Dir.pwd)}\/.+:\d+: warning: /)
+      .to_stderr
+  end
+end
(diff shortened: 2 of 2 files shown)


fornellas:
It failed again with 1.8.7... I can fix again, but does it make any sense to keep supporting a version that [has been obsoleted](https://www.ruby-lang.org/en/news/2013/06/30/we-retire-1-8-7/) in June 2013?

bf4:
@fornellas I have some code you can just dump in there and take credit :) https://github.com/metricfu/metric_fu/blob/master/spec/capture_warnings.rb

fornellas:
@bf4, your solution uses temporary files, which is slower than [RSpec's implementation](https://github.com/rspec/rspec-expectations/blob/af56605b6fb3b854830d527f5331c5c022c7e62d/lib/rspec/matchers/built_in/output.rb#L158) that uses StringIO. Other than that, it uses abut the same basic logic: process lines and fetch the ones relevant to project's code.

I recommend you pay attention to temporary files and dirs. It is usually recommended to prefer the block form:
```ruby
Tempfile.create('foo', '/home/temp') do |f|
   ... do something with f ...
end
```
because it ensures the file will be cleaned up, even if the block raises an exception. The same applies to $stderr.reopen. You should use a capture block, to ensure it is always restored to its original value (which you should also save). Example:
```ruby
require 'tempfile'
Tempfile.create('stderr') do |stderr_file|
  original_stderr = STDERR.clone
  begin
      STDERR.reopen(stderr_file)
      STDERR.puts "Example!!"
  ensure
    STDERR.reopen(original_stderr)
  end
  stderr_file.close
  puts "Captured via stderr file:"
  puts File.open(stderr_file.path).read
end
```

PragTob:
Ruby 1.8.7 isn't "really" supported only so far as that projects don't crash with it. It's the explicit goal of the project to remove that compatibility. Although I feel the pain ;) One might just skip the warning checks for ruby 1.8.7 or replace keep_if which might be rather easy.

PragTob:
Hi there,

thank you for the PR. I've replaced it with #556 that took the code Benjamin suggested and fiddled with it some.

Main reason being, that I didn't feel comfortable replacing Kernel before every spec and then restoring it after keeping the local variable in scope. The other solution is also more general and able to catch a wider range of errors (as you noted they come from more than one source).

On the note of performance, constantly replacing Kernel.warn makes it slower (at least in my testing) than using a Tempfil, I switched it over to StringIO nonetheless :)

About your `$stderr` concern, afaik STDERR is its original value so it's fine. Resetting it/using Tempfile in block form is quite a bit harder as it executes in a different scope/is called upon during rspec execution.

If you wanna review the PR I made, I'd be happy.

Thanks + cheers,
Tobi

PR #460, opened 2016-02-07
Update badges to SVG versions

craiglittle:
I think this gives the `README` a bit cleaner and more consistent look.


colszowka:
Cool, thanks!

PR #461, opened 2016-02-10
Unfreeze SimpleCov::VERSION

leafle:
Older versions of rubygems choke on an immutable version string.

I wasn't able to actually get rubocop to complain about the mutable constant so I couldn't verify that the second commit actually does anything.


colszowka:
Thanks!

PR #469, opened 2016-03-07
Don't overwrite last_run file if refuse_coverage_drop option is enabled and the coverage has dropped

Miloshes:
Refuse_coverage_drop option is great but if the coverage actually dropped, it will only fail your test suite once. The .last_run.json file gets overwritten and tests are passing again.

Proposed changes look for "refuse_coverage_drop" option and if it is enabled, prevent overriding of the file when the coverage has dropped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ Unreleased ([changes](https://github.com/colszowka/simplecov/compare/v0.11.2...m
 
 ## Enhancements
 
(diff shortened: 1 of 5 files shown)


PragTob:
Hey there, thanks for the PR! This seems like a problem indeed! I'll recheck and try to merge this. I'll drop e0a14d7 though as it was fixed in the mean time. Sorry for the longer down time here.

PragTob:
Rebased and reopened as #553

PR #473, opened 2016-03-22
Correct the detection and guide for JRuby coverage

donv:
`JRuby.runtime.debug?` does not answer correctly wether coverage is active for JRuby.  `org.jruby.RubyInstanceConfig.FULL_TRACE_ENABLED` does.

Enabling general debug mode on JRuby does not enable coverage.  You need the "--debug" flag on the command line or in JRUBY_OPTS, or set the "debug.fullTrace=true" option in ".jrubyrc".

This change should make using simplecov on JRuby a bit nicer.

### DIFF
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -10,9 +10,10 @@ if defined?(JRUBY_VERSION) && defined?(JRuby)
   # @see https://github.com/colszowka/simplecov/issues/86
   # @see https://jira.codehaus.org/browse/JRUBY-6106
 
-  unless JRuby.runtime.debug?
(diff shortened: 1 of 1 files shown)


bf4:
Cc @enebo

enebo:
@bf4 @donv Yeah I think this is as good as it gets right now.  I would like in the future to make this a bit less internals and expose it via some more "official" API but this will not change and it is more accurate way of detecting this condition for now.

donv:
@bf4 @xaviershay Good to merge?

donv:
Thanks!

PR #475, opened 2016-03-23
Limit rake version to 10.x for Ruby 1.8.7

### LINKED ISSUE #474: Resolve (travis-ci) test failures with MRI 1.8.7 vs. rake 11.x
Rake 11+ requires Ruby 1.9.3 or later.  Simplecov tests fail for MRI 1.8.7 when trying to install rake 11+.

For travis-ci, we need to either limit the rake version to '<11' or remove support for MRI 1.8.7.

I can help with either.

donv:
Fixes #474

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ source "https://rubygems.org"
 # Uncomment this to use development version of html formatter from github
 # gem 'simplecov-html', :github => 'colszowka/simplecov-html'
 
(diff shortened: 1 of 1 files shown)


xaviershay:
checked that we don't commit our `Gemfile.lock` so this seems sane to me.

Really should probably just drop 1.8.7, but this'll do for now.

PR #476, opened 2016-03-23
Avoid forking test for JRuby since fork is not implemented

donv:
This fixes 10 test failures for JRuby in travis

### DIFF
--- a/spec/faked_project/spec/forking_spec.rb
+++ b/spec/faked_project/spec/forking_spec.rb
@@ -2,6 +2,7 @@ require "spec_helper"
 
 describe "forking" do
   it do
-    Process.waitpid(Kernel.fork {})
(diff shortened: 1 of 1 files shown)


xaviershay:
seems this changed a magic coverage constants in the others tests :(

LGTM once those are fixed.

donv:
@xaviershay I assume that since I added new lines to the source, and they are covered, the percentage increased.  I'll adjust those, then.

donv:
@xaviershay Simplified the test change to avoid changing the coverage percentage.

This fixes all but one test for JRuby, and that was failing before this PR.  I will look at that separately from this PR.

Good to merge?

colszowka:
LGTM, merging. Thank you!

PR #477, opened 2016-03-24
Set useful JRUBY_OPTS

donv:
`--dev` sets useful options for faster startup and better performance for short-running apps, including -X-C
`--debug` is necessary on JRuby 9K to get coverage at all
`--1.9` is default for JRuby 1.7.x and not recognized on JRuby 9K


colszowka:
Thank you!

Issue #478, opened 2016-03-24
JRuby includes coverage for .simplecov


donv:
Opening this for guidance and discussion.  If you have a chat channel, I'd prefer that :)

I am trying to solve the last JRuby failure:

https://travis-ci.org/colszowka/simplecov/jobs/118230188#L234

JRuby includes `.simplecov` in the list of covered files.  Is that wrong?  It only happens for `Scenario: Using configure block after start`, and code (`SimpleCov.configure`) in `.simplecov` is run after the `SimpleCov.start`call has completed.

Seems like including `.simplecov` in the list of covered files could be considered correct behavior.

I can try setting up the test to have different expectations for JRuby and others.  What do you think?

colszowka:
Hmm, that is interesting. Technically, I guess since it is a ruby source file that is being loaded, one might want to have coverage applied, but on the other hand I think we should try to keep things similar between ruby implementations, and MRI Coverage does not seem to pay attention to this file (most likely because we use [load](https://github.com/colszowka/simplecov/blob/master/lib/simplecov/defaults.rb#L102) instead of require, and if I remember correctly MRI coverage only is supposed to track `require`d files).

I would suggest to add a jruby-specific filter for either `.simplecov` or (probably preferrably) `!EXTNAME == .rb` to https://github.com/colszowka/simplecov/blob/master/lib/simplecov/defaults.rb - maybe as a separate adapter like the `root_filter`, which then gets a conditional load in https://github.com/colszowka/simplecov/blob/master/lib/simplecov/defaults.rb#L46

donv:
I checked the doc, and it explicitly states that loaded code should be covered:

http://ruby-doc.org/stdlib-2.3.0/libdoc/coverage/rdoc/Coverage.html

Submitted a PR #479 for conditional expectation for JRuby.

I am willing to go another way if you still want that.

PragTob:
With #547 we probably don't need this anymore, more reasoning in #479 still lots of thanks :tada:

PR #479, opened 2016-03-24
Expect coverage for `.simplecov` when running with JRuby

donv:
...when running code in `.simplecov` after `SimpleCov.start` has finished.

Downside is we get a skipped scenario indicated with dashes.

### DIFF
--- a/features/config_styles.feature
+++ b/features/config_styles.feature
@@ -76,8 +76,24 @@ Feature:
     Then I should see "4 files in total."
     And I should see "using Config Test Runner" within "#footer"
 
+  Scenario: Using configure block after start with JRuby
+    Given the Ruby engine is jruby
+      And a file named ".simplecov" with:
+      """
+      SimpleCov.start
+      SimpleCov.configure do
+        add_filter 'test'
+        command_name 'Config Test Runner'
+      end
+      """
+
+    When I open the coverage report generated with `bundle exec rake test`
+    Then I should see "5 files in total."
+    And I should see "using Config Test Runner" within "#footer"
+
   Scenario: Using configure block after start
-    Given a file named ".simplecov" with:
+    Given the Ruby engine is not jruby
+      And a file named ".simplecov" with:
       """
       SimpleCov.start
       SimpleCov.configure do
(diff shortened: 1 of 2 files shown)


donv:
JRuby tests are green in travis!  :cake:

donv:
@colszowka Good to merge?

donv:
@colszowka @xaviershay @bf4 Anything more I need to do before merge?

xaviershay:
Thanks for looking at this!

Could you add some context for why we expect these to be different? As written the scenarios are pretty confusing - why do we "expect" JRuby to have different results!?

The answer is likely - coz that's just how it works on JRuby and we don't care enough to fix it? If true, we should either a) document that, b) not specify the behaviour (i.e. existing example is "over specified" since we don't care about number of files, or c) exclude JRuby with a comment that this behaviour is undefined on JRuby.

(Option b makes the most sense to me, but I haven't thought about the problem a lot)

donv:
Thanks for responding!

There is some discussion in #478 .  As you suggested, we expect this because that's how it is.

I confronted the JRuby team, and we feel we are doing the right thing.  The natural next step is filing an issue with https://github.com/ruby/spec with a test case to clear up the expected behavior.  I can do that (hopefully with some help from you?).

Before that, the behavior does differ across Ruby implementations.  Too bad ruby-head, ruby-2.3, and rbx don't have green tests for simplecov.  That could shed some light.  Anyway, I _really_ want to have green simplecov tests for JRuby before going further.  When the verdict from ruby/spec comes back, we can make changes where appropriate.

bf4:
And https://github.com/colszowka/simplecov/blob/783eb971986e34d85e41b9ff96ae9df1e4d9c154/lib/simplecov/jruby_fix.rb

xaviershay:
seems legit - can we put that comment in the spec (or at least in the commit message) so it's easy for future spelunkers to find?

donv:
OK, started writing some specs for coverage in ruby/spec, and it will lead to changes in JRuby at least.  So we can hold off this PR a little while longer.

PragTob:
JRuby 9.1 seems to have resolved failures and runs cleanly now and supporting this through #547 I think this one can get closed as we don't need conditional tests then.

If anyone ( ping @donv ) disagrees please reopen/let me know :)

Issue #470, opened 2016-03-09
couldn't find HOME environment -- expanding `~' (ArgumentError)


mhenrixon:
Hi,

You have the same issue as rubocop did in https://github.com/bbatsov/rubocop/issues/2901 simplecov crash unless a HOME environment variable is defined.

```
/var/lib/gems/2.1.0/gems/simplecov-0.11.2/lib/simplecov/defaults.rb:99:in `expand_path': couldn't find HOME environment -- expanding `~' (ArgumentError)
	from /var/lib/gems/2.1.0/gems/simplecov-0.11.2/lib/simplecov/defaults.rb:99:in `<top (required)>'
	from /var/lib/gems/2.1.0/gems/simplecov-0.11.2/lib/simplecov.rb:180:in `require'
	from /var/lib/gems/2.1.0/gems/simplecov-0.11.2/lib/simplecov.rb:180:in `<top (required)>'
	from /root/.jenkins/jobs/e24-accounting/workspace/spec/spec_helper.rb:5:in `require'
	from /root/.jenkins/jobs/e24-accounting/workspace/spec/spec_helper.rb:5:in `<top (required)>'
	from /root/.jenkins/jobs/e24-accounting/workspace/spec/features/bitcoin_deposit_obsoletes_deposit_address_spec.rb:1:in `require'
	from /root/.jenkins/jobs/e24-accounting/workspace/spec/features/bitcoin_deposit_obsoletes_deposit_address_spec.rb:1:in `<top (required)>'
	from /var/lib/gems/2.1.0/gems/rspec-core-3.4.1/lib/rspec/core/configuration.rb:1361:in `load'
	from /var/lib/gems/2.1.0/gems/rspec-core-3.4.1/lib/rspec/core/configuration.rb:1361:in `block in load_spec_files'
	from /var/lib/gems/2.1.0/gems/rspec-core-3.4.1/lib/rspec/core/configuration.rb:1359:in `each'
	from /var/lib/gems/2.1.0/gems/rspec-core-3.4.1/lib/rspec/core/configuration.rb:1359:in `load_spec_files'
	from /var/lib/gems/2.1.0/gems/rspec-core-3.4.1/lib/rspec/core/runner.rb:102:in `setup'
	from /var/lib/gems/2.1.0/gems/rspec-core-3.4.1/lib/rspec/core/runner.rb:88:in `run'
[...]

PR #482, opened 2016-04-04
Fix couldn't find HOME

waldyr:
#470


PragTob:
Thanks!

This is weird... I can't use the instructions to fetch the patch... This might get a bit wild. Hold on tight :)

PR #483, opened 2016-04-05
Update README.md

waldyr:


README.md
```diff
@@ -55,8 +55,8 @@ Getting started
 
     ```ruby
     require 'simplecov'
-    SimpleCov.start
-
+    SimpleCov.start # Or SimpleCov.start 'rails'
+
```
bf4: Would you mind providing a little background on this change?

waldyr: The start of simplecov with rails built-in configuration is a little hard to find in the Readme. We should disclose it in a different fashion.

PragTob:
Hi there,

at least these days `SimpleCov.start 'rails'` is mentioned a lot in the README so I think we should be good. If you disagree strongly please let me know :)

Anyhow, still thanks for the PR :)

PR #488, opened 2016-04-12
use around

grosser:
@sinsoku

### DIFF
--- a/spec/return_codes_spec.rb
+++ b/spec/return_codes_spec.rb
@@ -4,14 +4,11 @@ require "helper"
 # See https://github.com/colszowka/simplecov/issues/5
 describe "return codes" do
   context "inside fixtures/frameworks" do
-    before do
-      @current_dir = Dir.getwd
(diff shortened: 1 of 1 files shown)


grosser:
@xaviershay @colszowka I can haz merge ?

spec/return_codes_spec.rb
```diff
@@ -4,14 +4,8 @@
 # See https://github.com/colszowka/simplecov/issues/5
 describe "return codes" do
   context "inside fixtures/frameworks" do
-    before do
-      @current_dir = Dir.getwd
-      Dir.chdir(File.join(File.dirname(__FILE__), "fixtures", "frameworks"))
-      FileUtils.rm_rf("./coverage")
```
xaviershay: why is it ok to drop this line?

grosser: it is not, fixed!

xaviershay:
great, thanks

PR #491, opened 2016-04-13
remove json runtime dependency

### LINKED ISSUE #458: dropping json gem dependency
hello simplecov-contributors!

just out of interest, how important is supporting really really really old rubies for simplecov?
i recently saw this tweet from Mike Perham (https://github.com/rails/rails/pull/23453) and started looking at our Gemfiles. some of the projects only use simplecov and therefor a quite significant amount of gem dependencies.

dropping the json is just one example and it's only a minor code change, it's almost a philosophical question and there is no right or wrong. there are opinions. i'm interested in the opinion of you, the simplecov contributors.

thanks for simplecov!
ben

BenMorganIO:
Similar to https://github.com/rails/rails/pull/23453 and fixes #458.

### DIFF
--- a/simplecov.gemspec
+++ b/simplecov.gemspec
@@ -14,7 +14,6 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = ">= 1.8.7"
 
-  gem.add_dependency "json", "~> 1.8"
   gem.add_dependency "simplecov-html", "~> 0.10.0"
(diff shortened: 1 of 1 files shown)


BenMorganIO:
Specs are failing due to Ruby 1.8.7. We could drop support for Ruby 1.8.7 (that may upset some people) or perhaps we could do a detection for the json gem and warn them that they need to install the gem themselves?

PR #492, opened 2016-04-13
add channels group to rails profile

BenMorganIO:
This basically adds Channels as a group for test coverage. What I don't like about this is that it assumes (a) you're on Rails 5 and (b) that you're using ActionCable. I'm very open to modifying this PR to support Rails <5 and just do a detection for ActionCable.

### DIFF
--- a/lib/simplecov/defaults.rb
+++ b/lib/simplecov/defaults.rb
@@ -29,6 +29,7 @@ SimpleCov.profiles.define "rails" do
   add_filter "/db/"
 
   add_group "Controllers", "app/controllers"
+  add_group "Channels", "app/channels"
   add_group "Models", "app/models"
   add_group "Mailers", "app/mailers"
   add_group "Helpers", "app/helpers"


dtaniwaki:
👍

lib/simplecov/defaults.rb
```diff
@@ -29,6 +29,7 @@
   add_filter "/db/"
 
   add_group "Controllers", "app/controllers"
+  add_group "Channels", "app/channels"
```
bf4: could be something like  `  add_group "Channels", "app/channels" if defined?(ActionCable) || defined?(Rails) && Rails::VERSION::MAJOR >= 5` but the below might be cleaner.   

```ruby
SimpleCov.profiles.define "rails5" do
  load_profile "rails"

  add_group "Channels", "app/channels"
end
```
 
@colszowka @xaviershay Any preference for adding the group here in all Rails version, with a condition, or making a new rails5 profile?

BenMorganIO: Maybe just check for ActionCable. I know that in Rails 5 APIs, ActionCable isn't exactly used much :p

PragTob:
Hi there,

thanks for the PR! I think as the others suggested just adding it when ActionCable is defined might be a good idea. 

If you have time, please go ahead and do it, otherwise I can also take over :)

PragTob:
Merged as 5571d0d8e00744d :)

Thanks!

PR #506, opened 2016-06-17
Fix merging of resultset 

hanazuki:
In merging two resultsets, if a file only appears in only one of the resultsets, the current merger implementation generates incorrect resultset.

- Merging `{file => [nil, 0, 1]}` and `{}` results in `{file => [nil, 0, 1]}` (ok)
- Merging `{}` and `{file => [nil, 0, 1]}` results in `{file => [nil, nil, 1]}` (0 is lost)

This patch fixes the problem.

### DIFF
--- a/lib/simplecov/merge_helpers.rb
+++ b/lib/simplecov/merge_helpers.rb
@@ -22,11 +22,14 @@ module SimpleCov
     def merge_resultset(hash)
       new_resultset = {}
       (keys + hash.keys).each do |filename|
-        new_resultset[filename] = []
+        new_resultset[filename] = nil
       end
 
       new_resultset.each_key do |filename|
-        new_resultset[filename] = (self[filename] || []).extend(SimpleCov::ArrayMergeHelper).merge_resultset(hash[filename] || [])
+        result1 = self[filename]
+        result2 = hash[filename]
+        new_resultset[filename] =
+          (result1 && result2) ? result1.extend(ArrayMergeHelper).merge_resultset(result2) : (result1 || result2).dup
       end
       new_resultset
     end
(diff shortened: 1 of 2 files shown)


bf4:
possibly related https://github.com/colszowka/simplecov/issues/501#issuecomment-226802034

ayufan:
Actually there's one more issue with merging.

It's quite common that you have:

`{file => [1, nil, nil]}` to merge with `{file => [0, 0, 0]}` results in `{file => [1, 0, 0]}` which is incorrect.

SimpleCov for file that is only loaded, but not executed prefills coverage with zero's. The `nil` has a special meaning here: this line were processed, but ignored. This results in badly calculated coverage report when you have specs that touches different files and their result is then merged. The ignored lines are marked as not covered, where they should be ignored.

Given above. Proper result for this is: `{file => [1, nil, nil]}`.

hanazuki:
@ayufan 
That case is fixed in #441 (not yet released) and now it seems working correctly.

```console
% git rev-parse @
a80baea9eb97772eeeec339d379887368969c91c
% bundle exec irb -rsimplecov
irb(main):001:0> [1,nil,nil].extend(SimpleCov::ArrayMergeHelper).merge_resultset([0,0,0])
=> [1, nil, nil] (size=3)
irb(main):002:0> [0,0,0].extend(SimpleCov::ArrayMergeHelper).merge_resultset([1,nil,nil])
=> [1, nil, nil] (size=3)
```

ayufan:
Awesome!

sferik:
Working on a new release now!

thedrow:
@hanazuki @sferik We should close this one then right?

sferik:
@thedrow Yes, I believe so. Please re-open it if this issue reappears.

hanazuki:
@thedrow @sferik 
No, this problem is not fixed yet.
Actually this one is a regression of #441, which intoroduces nil-awareness to the merger procedure.

PR #512, opened 2016-07-09
Require json gem only if Ruby < 2

krzysiek1507:


krzysiek1507:
@sferik

bf4:
Closing as this is not supported by rubygems. The gemspec is evaluated at the time the gem is built.

PR #513, opened 2016-07-13
Fix merging of result sets for unloaded files

hanazuki:
reopens #506 

- Merging `{file => [nil, 0, 1]}` and `{}` results in `{file => [nil, 0, 1]}` (ok)
- Merging `{}` and `{file => [nil, 0, 1]}` results in `{file => [nil, nil, 1]}` (0 is lost, this is the bug)

When merging two result sets, if a file is missing on the LHS, which means the file is not loaded, zeros for the file on the RHS will be lost.

### DIFF
--- a/lib/simplecov/merge_helpers.rb
+++ b/lib/simplecov/merge_helpers.rb
@@ -22,11 +22,14 @@ module SimpleCov
     def merge_resultset(hash)
       new_resultset = {}
       (keys + hash.keys).each do |filename|
-        new_resultset[filename] = []
+        new_resultset[filename] = nil
       end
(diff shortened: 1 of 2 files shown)


sferik:
@hanazuki Thanks!

yujinakayama:
@sferik Could you cut a patch release for this please?

MikeMcQuaid:
@sferik Would also be ❤️ for Homebrew to have a patch release for this (and https://github.com/colszowka/simplecov/pull/520). Let me know if I can help at all! Thanks for the awesome project 🎉

PR #520, opened 2016-09-21
Fix an accidentally inefficient hash filter

penman:
I (accidentally ⏤ Homebrew/brew#1041) loaded a very large number of files into SimpleCov, and it got stuck for a very long time. I tracked down the slowness to this method, which gets very slow because it iterates `filenames` for every value of `original_result`. Using Ruby's built-in `Array#&` makes this much faster.


colszowka:
Neat, thank you!

PR #529, opened 2016-10-19
Test against latest Ruby 2.3 release

lucianosousa:
it's 2.3.1, already available on travis


colszowka:
Thanks, but Travis didn't like it:

```
4.83s$ rvm use 2.3 --install --binary --fuzzy
ruby-2.3 is not installed - installing.
Searching for binary rubies, this might take some time.
Requested binary installation but no rubies are available to download, consider skipping --binary flag.
Gemset '' does not exist, 'rvm ruby-2.3 do rvm gemset create ' first, or append '--create'.
The command "rvm use 2.3 --install --binary --fuzzy" failed and exited with 2 during .
```

I guess you'd have to bump it to 2.3.1 explicitly here

colszowka:
Thanks!

PR #530, opened 2016-10-20
Allow coverage directory to be a symlink.

johann-p-koebbe:
In a Vagrant setup, I have symlinked the coverage directory to a shared folder so I can easily access it from the host. In 0.12, I get an error when the symlink exists and simplecov tries to create the directory. This change allows for the possibility that coverage is a symlink. I've also added some specs around creating the directory and such.


johann-p-koebbe:
I'm still seeing some wonky behavior, so I don't think this PR is ready for review.

PR #531, opened 2016-10-21
Skip the :nocov: comments.

ebiven:
When using the `nocov` comment token the start and stop tokens are considered lines to be covered. Addresses the following issues:

https://github.com/colszowka/simplecov/issues/468

https://github.com/colszowka/simplecov/issues/527


PragTob:
Hi there!

Thanks for the PR :star: 

It seems that this expectation is ingrained into the test suite although I agree it shouldn't be a line to be covered :D So, do you want to fix up the tests or shall I?

ebiven:
I'll get those this morning and update the pr. 

Eric

ebiven:
Moved to #551

PR #540, opened 2016-12-13
Stop to read unused source code

yui-knk:
Before this commit all source code were loaded, even if
they are not used.
For example when taking test coverage for rails applications,
in many cases files on "vendor/bundle" are filtered out.
By reading source code lazily, the time required to generate
results will be shortened.


yui-knk:
@colszowka I added comment why we read lazily 😄

PragTob:
Looks good to me :+1:

colszowka:
:+1: Thank you @yui-knk !

amatsuda:
👏

yui-knk:
😄

PR #547, opened 2017-01-25
Officially Support JRuby 9.1+

### LINKED ISSUE #524: Status of JRuby Support
What is the status of support for the various flavors of Ruby, in particular JRuby? Is JRuby 9000 supported? What about JRuby 1.7? The "Ruby version compatibility" section of the README has not been updated [since 2011](https://github.com/colszowka/simplecov/commit/1b13859b7fb44102de14e04658e12adae7eced21#diff-cb3e0f2c76a671c083e8f001970f4631).

PragTob:
I've personally been running simplecov with JRuby 1.7+ for years
so I'd say it is good to go either way ;P

The reality of edge cases is more complicated but it seems the
fine folks at @jruby (namely @enebo ) made the last fixes
necessary as seen in jruby/jruby#1196 so that they're available
in 9.1+ releases! The others should work, but have some little
failures (9.0.5.0 fails on one cuke for me).

So, the easiest way forward is to support 9.1+. Sadly the tests
take a while as they always start new interpreter instances and
startup time isn't exactly JRuby's strength, but CI does it
so it's bearable. :)

If anyone has problems with JRuby on a PR, in an issue or whatever
please feel free to ping me (@PragTob)

Also, special thanks goes to @donv who created did a lot of testing,
issue reporting et. al. for simplecov on JRuby for a long time!

* fixes #524

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,5 @@ matrix:
   allow_failures:
     - rvm: ruby-head
     - rvm: jruby-head
-    - rvm: jruby-9.1.7.0
     - rvm: rbx-2
   fast_finish: true
--- a/README.md
+++ b/README.md
@@ -582,14 +582,11 @@ SimpleCov.formatters = [
 
 ## Ruby version compatibility
 
-Only Ruby 1.9+ ships with the coverage library that SimpleCov depends upon.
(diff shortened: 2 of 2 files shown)


colszowka:
:fireworks: !

PR #551, opened 2017-01-26
Skip the :nocov: comments.

ebiven:
When using the `nocov` comment token the start and stop tokens are considered lines to be covered. Addresses the following issues:

colszowka#468

colszowka#527


ebiven:
This is a refresh of #531 since the repo had moved for three months.

PragTob:
Looks good, all test changes make sense to me (e.g. from never to skipped).

Thanks a lot!

PR #553, opened 2017-01-29
Don't overwrite last_run file if refuse_coverage_drop option is enabled and the coverage has dropped 

PragTob:
Rebased/slightly cut reopen of #469 by @Miloshes

Original text:

> Refuse_coverage_drop option is great but if the coverage actually dropped, it will only fail your test suite once. The .last_run.json file gets overwritten and tests are passing again.
>
> Proposed changes look for "refuse_coverage_drop" option and if it is enabled, prevent overriding of the file when the coverage has dropped.

### DIFF
--- a/lib/simplecov/defaults.rb
+++ b/lib/simplecov/defaults.rb
@@ -70,6 +70,7 @@ at_exit do # rubocop:disable Metrics/BlockLength
   if SimpleCov.result? # Result has been computed
     covered_percent = SimpleCov.result.covered_percent.round(2)
     covered_percentages = SimpleCov.result.covered_percentages.map { |p| p.round(2) }
+    coverage_diff = 0
 
     if @exit_status == SimpleCov::ExitCodes::SUCCESS # No other errors
       if covered_percent < SimpleCov.minimum_coverage # rubocop:disable Metrics/BlockNesting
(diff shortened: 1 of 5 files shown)


lib/simplecov/defaults.rb
```diff
@@ -79,15 +80,18 @@
...
+        if coverage_diff > SimpleCov.maximum_coverage_drop # rubocop:disable Metrics/BlockNesting
+          $stderr.printf("Coverage has dropped by %.2f%% since the last time (maximum allowed: %.2f%%).\n", coverage_diff, SimpleCov.maximum_coverage_drop)
           @exit_status = SimpleCov::ExitCodes::MAXIMUM_COVERAGE_DROP
         end
       end
     end
 
-    SimpleCov::LastRun.write(:result => {:covered_percent => covered_percent})
+    # Don't overwrite last_run file if refuse_coverage_drop option is enabled and the coverage has dropped
+    unless SimpleCov.refuse_coverage_drop && coverage_diff > 0
```
PragTob: If I understand this code correctly, then the call to `SimpleCov.refuse_coverage_drop` will refuse a coverage drop. If I'm right we shouldn't do this and if the tests really don't catch this I should add a test :D

Miloshes:
Thanks @PragTob

PR #555, opened 2017-01-29
Use e.g. instead of i.e.

restlessdesign:
“i.e.” means “that is” or “in other words” and is mean to clarify, whereas “e.g.” means “for example”, which is what these instances in the documentation are providing (i.e., examples) :)


PragTob:
Thanks!

PR #556, opened 2017-02-04
Fails specs on ruby warnings

PragTob:
Originally stolen from: https://github.com/metricfu/metric_fu/blob/master/spec/capture_warnings.rb

Heavily modified though - e.g. put into a class, using stringio,
removed the bundler directory as not used here and moved to
before the before of the rspec suite to also capture parse time
warning errors.

Replaces #457 ( cc @bf4 @fornellas ) - basically decided for this as it doesn't mess with `Kernel.warn` and is able to capture more, even rearranged it so that it catches parse time errors :tada: 

As this is some functionality I've wanted for a long time, I heavily consider extracting this and making it a small little gem that could then also play with Cucumber/Minitest and friends :)

### DIFF
--- a/spec/helper.rb
+++ b/spec/helper.rb
@@ -1,5 +1,7 @@
-require "simplecov"
 require "rspec"
+# loaded before simplecov to also capture parse time warnings
+require "support/fail_rspec_on_ruby_warning"
+require "simplecov"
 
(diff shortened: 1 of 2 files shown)


zenspider:
I would love to have this extracted and available for regular testing.

PR #558, opened 2017-02-14
Rewrite merge helpers to work with frozen objects

aroben:
`Coverage.result` returns a frozen Hash containing frozen Strings as keys and frozen Arrays as values. The existing merge helper did not work when passed a `Coverage.result` directly, because it would try to `extend` the frozen Arrays. This is not a problem in typical SimpleCov usage because usually Coverage.result has been roundtripped through `.resultset.json` resulting in an unfrozen copy. But when trying to do manual merging of `Coverage.result` hashes the use of `extend` gets in the way.

The merge helper has now been rewritten a method that takes two Hashes and doesn't modify them or their contents at all. While I was touching this code I tried to clean it up a little bit as well.


lib/simplecov/result.rb
```diff
@@ -24,7 +24,6 @@ class Result
     # Initialize a new SimpleCov::Result from given Coverage.result (a Hash of filenames each containing an array of
     # coverage data)
     def initialize(original_result)
-      original_result = original_result.dup.extend(SimpleCov::HashMergeHelper) unless original_result.is_a? SimpleCov::HashMergeHelper
```
bf4: 👍

lib/simplecov/merge_helper.rb
```diff
@@ -0,0 +1,31 @@
+module SimpleCov
+  module MergeHelper
+    class << self
+      # Merges two Coverage.result hashes
+      def merge_resultsets(a, b)
```
bf4: I think this is a long needed feature. Any reason it shouldn't just be a function on the ResultMerger?

lib/simplecov/result_merger.rb
```diff
@@ -62,7 +62,7 @@ def results
       def merged_result
         merged = {}
         results.each do |result|
-          merged = result.original_result.merge_resultset(merged)
+          merged = SimpleCov::MergeHelper.merge_resultsets(result.original_result, merged)
```
bf4: wouldn't it be nice if this were `merged = SimpleCov::RawCoverage.merge_results(results.map(&:original_result))`

```ruby
module SimpleCov::RawCoverage
  module_function

  def merge_results(*results)
    Array(results).reduce({}) do |result, merged|
      merge_resultset(result, merged)
    end
  end

  def merge_resultset(resultset1, resultset2)
    (resultset1.keys | resultset2.keys).each_with_object({}) do |filename, merged_resultset|
      coverage1 = resultset1[filename]
      coverage2 = resultset2[filename]
      merged_resultset[filename] = merge_file_coverage(coverage1, coverage2)
    end
  end

  def merge_file_coverage(coverage1, coverage2)
    if coverage1 && coverage2
      coverage1.zip(coverage2).map { |count1, count2|
        sum = count1.to_i + count2.to_i
        if sum.zero? && (count1.nil? || count2.nil?)
          nil
        else
          sum
        end
      }
    else
      (coverage1 || coverage2).dup
    end
  end
end
```

lib/simplecov/merge_helper.rb
aroben: Putting in on ResultMerger seems sensible. I considered it, then thought maybe it'd be nicer to keep this code self-contained. I.e., ResultMerger manages `.resultset.json`, and this module manages merging results in-memory. But it was a close call, and if you prefer moving it to ResultMerger I'll do that.

lib/simplecov/result_merger.rb
aroben: I had actually been thinking about adding a method that takes an array of `SimpleCov::Result` and merges them. ResultMerger does this internally but doesn't expose it as an API. That could certainly build on the API you've proposed here.

Earlier you mentioned putting this code in ResultMerger, but now it sounds like you'd like it in a new `RawCoverage` module instead? Happy to do whatever seems best to you.

bf4: I don't have a strong feeling about it except that this something I've been meaning to do or hoping someone would do for months so 👍 to you!  Only real concern is that we want to make a small public interface so that we can change internal details. So, maybe a method on ResultMerger that calls the helper or whatever, and we can change the whatever however whenever :)

aroben:
@bf4 OK I added the new APIs. I didn't add a specific test for `ResultMerger.merge_results` because it felt like it would duplicate the tests for `ResultMerger.merged_result` quite a bit. But I'd be happy to add some if you think they'd be worthwhile.

lib/simplecov/result_merger.rb
```diff
@@ -54,20 +54,24 @@ def results
         results
       end
 
+      # Merge two or more SimpleCov::Results into a new one with merged
+      # coverage data and the command_name for the result consisting of a join
+      # on all source result's names
+      def merge_results(*results)
```
bf4: 👍

bf4: Since there's already a `results` method and it is passed to this method as the argument, perhaps we can avoid possible which-results-are-we-using by changing the argument name? `*resultsets`? maybe?

aroben: I think you're right that there could be some confusion. I'm a little wary of calling this `resultsets` because this is an array of `SimpleCov::Result`s, not an array of `.resultset.json`-type data. That actually seems like a problem with `RawCoverage.merge_resultsets` too. But I'm not super familiar with SimpleCov's terminology, so maybe `resultsets` is an OK name here?

Without any concern for backward compatibility, I'd be tempted to rename `ResultMerger.results` to `ResultMerger.stored_results`, and leave the argument here as `*results`.

aroben: @bf4 Let me know how you'd like to proceed here. ❤️

lib/simplecov/raw_coverage.rb
```diff
@@ -0,0 +1,38 @@
...
+    # Merges multiple Coverage.result hashes
+    def merge_results(*results)
+      results.reduce({}) do |result, merged|
+        merge_resultsets(result, merged)
+      end
+    end
+
+    # Merges two Coverage.result hashes
+    def merge_resultsets(result1, result2)
+      (result1.keys | result2.keys).each_with_object({}) do |filename, merged|
```
PragTob: I like the usage of `|` here

lib/simplecov/result_merger.rb
PragTob: I don't think I'd like to break backwards compatibility. I mean I wouldn't really consider it a public API of this gem, but from what I've seen I think some people have fiddled with this to merge results from multiple sources themselves so I wouldn't wanna break it for them.

spec/raw_coverage_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        }.freeze
+
+        @resultset2 = {
+          source_fixture("sample.rb").freeze => [1, nil, 1, 1, nil, nil, 1, 1, nil, nil].freeze,
+          source_fixture("app/models/user.rb").freeze => [nil, 1, 5, 1, nil, nil, 1, 0, nil, nil].freeze,
+        }.freeze
+      end
+
+      it "can merge without exceptions" do
+        SimpleCov::RawCoverage.merge_results(@resultset1, @resultset2)
```
PragTob: :+1: 

To clarify intent we could do `expect {}.not_to raise_error` or something, not strictly needed though

lib/simplecov/result_merger.rb
bf4: @PragTob I listed a few in https://github.com/colszowka/simplecov/pull/558#pullrequestreview-21775132

bf4: @PragTob @aroben how about

```ruby
def merge_results(*results = results())
```

this is something that I know has been done in Rails at Matz's recommendation when the default value for a param is a method by the same name

spec/raw_coverage_spec.rb
bf4: Yeah, I would normally do something at the top like

```ruby
example do
  expect {
     SimpleCov::RawCoverage.merge_results(@resultset1, @resultset2)
  }.not_to raise_error
end
```

I like `example` for the 'this is a valid usage', though I know some don't like it

lib/simplecov/raw_coverage.rb
```diff
@@ -0,0 +1,38 @@
...
+        file1 = result1[filename]
+        file2 = result2[filename]
+        merged[filename] = merge_file_coverage(file1, file2)
+      end
+    end
+
+    def merge_file_coverage(file1, file2)
+      return (file1 || file2).dup unless file1 && file2
+
+      file1.zip(file2).map do |count1, count2|
```
bf4: I wonder about zip vs. map.with_index

```ruby
GC.disable
a = Array.new(100) { 1 }.freeze; b = Array.new(100) { 2 }.freeze

count = GC.stat[:total_allocated_objects]; a.zip(b).map {|l,r| l + r }.tap { puts GC.stat[:total_allocated_objects] - count }
# 104
# => [3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3]

count = GC.stat[:total_allocated_objects]; a.map.with_index {|i,index| i + b[index] }.tap { puts GC.stat[:total_allocated_objects] - count }
# 5
# => [3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3


require 'benchmark/ips'
Benchmark.ips do |x| x.report('zip.map') { a.zip(b).map {|l,r| l + r } }; x.report('map.with_index') { a.map.with_index {|i,index| i + b[index] } }; x.compare! end
```

```
Warming up --------------------------------------
             zip.map     5.039k i/100ms
      map.with_index     7.615k i/100ms
             zip.map     40.456k (±10.8%) i/s -    201.560k in   5.040494s
      map.with_index     76.949k (± 2.9%) i/s -    388.365k in   5.051459s

Comparison:
      map.with_index:    76949.5 i/s
             zip.map:    40455.6 i/s - 1.90x  slower

```

bf4: looks like previous behavior was each_with_index

```diff
-    def merge_resultset(array)
 -      new_array = dup
 -      array.each_with_index do |element, i|
 -        pair = [element, new_array[i]]
 -        new_array[i] = if pair.any?(&:nil?) && pair.map(&:to_i).all?(&:zero?)
 -                         nil
 -                       else
 -                         element.to_i + new_array[i].to_i
 -                       end
 -      end
 -      new_array
 -    end
 -  end
```

PragTob: Good digging into that, thanks! :)

aroben: Looks like we can have the best of both worlds by using `lazy.zip.map`:

<details>
<summary><tt>bench.rb</tt></summary>

```ruby
require "benchmark/ips"

a = Array.new(100) { 1 }.freeze
b = Array.new(100) { 2 }.freeze

Benchmark.ips do |x|
  x.report('lazy.zip.map') { a.lazy.zip(b).map {|l,r| l + r } }
  x.report('zip.map') { a.zip(b).map {|l,r| l + r } }
  x.report('map.with_index') { a.map.with_index {|i,index| i + b[index] } }
  x.report('each_with_index') do |times|
    new_array = []
    i = 0
    while i < times
      a.each_with_index {|i,index| new_array << i + b[index] }
      i = i + 1
    end
  end
  x.compare!
end
```

</details>

```
Warming up --------------------------------------
        lazy.zip.map    17.413k i/100ms
             zip.map     7.960k i/100ms
      map.with_index    10.779k i/100ms
     each_with_index     9.897k i/100ms
        lazy.zip.map    186.025k (± 3.4%) i/s -    940.302k in   5.060879s
             zip.map     81.511k (± 4.0%) i/s -    413.920k in   5.087448s
      map.with_index    110.948k (± 3.4%) i/s -    560.508k in   5.058655s
     each_with_index    115.360k (± 2.2%) i/s -    583.923k in   5.064205s

Comparison:
        lazy.zip.map:   186024.9 i/s
     each_with_index:   115360.3 i/s - 1.61x  slower
      map.with_index:   110947.8 i/s - 1.68x  slower
             zip.map:    81510.6 i/s - 2.28x  slower
```

aroben: Whoops, that was so lazy that it didn't even create a result array. A correct implementation (with a `.to_a` on the end) is slower:

```
Warming up --------------------------------------
   lazy.zip.map.to_a     1.450k i/100ms
             zip.map     7.869k i/100ms
      map.with_index    10.748k i/100ms
     each_with_index     9.976k i/100ms
   lazy.zip.map.to_a     13.688k (±14.1%) i/s -     68.150k in   5.097768s
             zip.map     79.369k (± 8.6%) i/s -    393.450k in   5.005312s
      map.with_index    110.505k (± 4.0%) i/s -    558.896k in   5.067126s
     each_with_index    114.300k (± 4.2%) i/s -    578.608k in   5.071720s

Comparison:
     each_with_index:   114300.5 i/s
      map.with_index:   110505.2 i/s - same-ish: difference falls within error
             zip.map:    79368.7 i/s - 1.44x  slower
   lazy.zip.map.to_a:    13688.2 i/s - 8.35x  slower
```

spec/raw_coverage_spec.rb
aroben: I added `expect {}.not_to raise_error` to this test.

lib/simplecov/raw_coverage.rb
aroben: I've reworked this to use `map.with_index`. Thanks for the attention to performance!

bf4:
I think that rubocop `expect {` failure should be ignored in specs. @PragTob @xaviershay @colszowka  @sferik ?

https://travis-ci.org/colszowka/simplecov/jobs/202748911

```
Running RuboCop...
Inspecting 76 files
..........................................................................C.
Offenses:
spec/raw_coverage_spec.rb:89:16: C: Avoid using {...} for multi-line blocks.
        expect {
               ^
```

PragTob:
@bf4 if `do .. end` works for this case I'm all for using `do .. end` - but somewhere in the back of my mind I thought that might have caused problems somewhere along the way?

bf4:
@PragTob `expect do ... end` works, it's just not how people usually write it.

spec/raw_coverage_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+        @resultset2 = {
+          source_fixture("sample.rb").freeze => [1, nil, 1, 1, nil, nil, 1, 1, nil, nil].freeze,
+          source_fixture("app/models/user.rb").freeze => [nil, 1, 5, 1, nil, nil, 1, 0, nil, nil].freeze,
+        }.freeze
+      end
+
+      it "can merge without exceptions" do
+        expect {
+          SimpleCov::RawCoverage.merge_results(@resultset1, @resultset2)
+        }.not_to raise_error
```
bf4: Let's make this a positive assertion. 

```ruby
it "merges frozen resultsets" do
        resultset1 = {
          source_fixture("sample.rb").freeze => [nil, 1, 1, 1, nil, nil, 1, 1, nil, nil].freeze,
          source_fixture("app/models/user.rb").freeze => [nil, 1, 1, 1, nil, nil, 1, 0, nil, nil].freeze,
        }.freeze

        resultset2 = {
          source_fixture("sample.rb").freeze => [1, nil, 1, 1, nil, nil, 1, 1, nil, nil].freeze,
        }.freeze

        merged_result = SimpleCov::RawCoverage.merge_results(resultset1, resultset2)
        expect(merged_result.keys).to eq(resultset1.keys)
        expect(merged_result.values.map(&:frozen?)).to eq([false, false])
        expect(source_fixture('sample.rb')).to eq([whatever])
        expect(source_fixture('app/models/user.rb')).to eq([nil, 1, 5, 1, nil, nil, 1, 0, nil, nil])
       
end
```

aroben: OK, adopted that test and pushed.

PragTob:
RUbocop doesn't like single quotes - sorry :) 

```
spec/raw_coverage_spec.rb:89:43: C: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

      expect(merged_result[source_fixture('app/models/user.rb')]).to eq([nil, 1, 1, 1, nil, nil, 1, 0, nil, nil])
```

aroben:
Hm, we need to teach RuboCop to lint comments on GitHub :grin:. Fixed!​

spec/raw_coverage_spec.rb
bf4: why solve a problem when you can avoid it altogether? :)

bf4:
I'm trying this out.

PragTob:
@bf4 what do you mean by "trying this out" ? Like you run it locally on your projects to make sure it's good for an impeding release? :)

bf4:
Yup... It looks good... But since have a use case, I wanted to try it

B mobile phone

lib/simplecov/result.rb
```diff
@@ -24,7 +24,6 @@ class Result
...
       @original_result = original_result.freeze
       @files = SimpleCov::FileList.new(original_result.map do |filename, coverage|
```
bf4: Looks good. In my own testing I found that the `files` list wasn't set correctly if I was trying to load and merge the results from another machine.  Probably good for a future pr... but doesn't work with html formatter because it tries to read each file to overlay coverage in report  source_file.rb ``` def src
      # We intentionally read source code lazily to
      # suppress reading unused source code.
      @src ||= File.open(filename, "rb", &:readlines)
    end
```

```ruby
require 'simplecov'
require 'json'
SimpleCov.coverage_dir 'coveragetest'
SimpleCov.formatters = [SimpleCov::Formatter::HTMLFormatter]
sourcefiles = Dir.glob('./resultset*') 
# work around  `if File.file?(filename)` when running on other filesystem
add_files = ->(sr) { files = sr.original_result.map {|filename, coverage| SimpleCov::SourceFile.new(filename, coverage) }.compact.sort_by(&:filename); sr.instance_variable_set(:@files, SimpleCov::FileList.new(files)); sr }

results = sourcefiles.map { |file| SimpleCov::Result.from_hash(JSON.parse(File.read(file))) }.each do |result|
  add_files.(result)
end; nil
result = SimpleCov::ResultMerger.merge_results(*results)
add_files.(result)
result.format!
```

PragTob: so erm, I don't fully understand. Is this good to merge now or is this a bigger problem? (first you said it's probably good for another PR but then you said it doesn't work with the html formatter, so I'm a bit confused)

bf4: @PragTob good to merge. The 'good for another pr', would be a way to merge files not represented on the current file system, since that's out of scope. This works fine when running on the same filesystem (or if you were to munge the paths).

bf4:
@PragTob missing https://github.com/colszowka/simplecov/blob/master/CHANGELOG.md entry.  We can add it if @aroben isn't around

PragTob:
I usually add it on master after a merge so we should be fine :)

bf4:
@PragTob Then I'll leave it to you to bring this home :)

PragTob:
Sorry, a bit of life happened but merging this now. Thanks a lot @aroben for the great change and @bf4 for the great review :)

Issue #559, opened 2017-02-16
I am experiencing race condition while using SimpleCov with Parallel_tests


aminariana:
Repro:
I'm using `simplecov (0.13.0)`, `ruby "2.3.3"`, `parallel_tests (2.13.0)`
```
# In spec_helper.rb:
if ENV['COVERAGE']
  require 'simplecov'
  process_name = "process##{Process.pid.to_s}:#{ENV['TEST_ENV_NUMBER'] || '1'}"
  puts "Starting Code Coverage in #{process_name}"
  SimpleCov.command_name process_name
  SimpleCov.merge_timeout 120 # Expire parallel process coverage cache after x seconds
  SimpleCov.start 'rails'
  SimpleCov.refuse_coverage_drop
  SimpleCov.maximum_coverage_drop 1
  SimpleCov.minimum_coverage 100
end
```
Ran:
```
COVERAGE=true RAILS_ENV=test bundle exec rake parallel:spec
```

Expected:
The last of the 8 processes reports 100% coverage, which should make SimpleCov pass.

Actual:
The result of the 3rd-last process was picked as the merged result!
```
...
Coverage report generated for process#1306:3, process#1311:7, process#1313:8, process#1317:4, process#1318:5, process#1319:6 to /Users/amin/api/coverage. 1713 / 1801 LOC (95.11%) covered.
Coverage report generated for process#1306:3, process#1309:, process#1311:7, process#1313:8, process#1315:2, process#1317:4, process#1318:5, process#1319:6 to /Users/amin/api/coverage. 1754 / 1801 LOC (97.39%) covered.
Coverage report generated for process#1306:3, process#1309:, process#1311:7, process#1313:8, process#1315:2, process#1317:4, process#1318:5, process#1319:6 to /Users/amin/api/coverage. 1801 / 1801 LOC (100.0%) covered.
Coverage (95.11%) is below the expected minimum coverage (100.00%).

287 examples, 0 failures

Took 35 seconds
Tests Failed
```

bf4:
Have you tried setting `SimpleCov.pid = Process.pid` before forking?

aminariana:
I did just now, and it didn't help. Also please notice that `SimpleCov.start` already does this for you:
https://github.com/colszowka/simplecov/blob/abdf6a294f030a9407606adcf541d52505622b1b/lib/simplecov.rb#L47

PragTob:
[...]

PR #570, opened 2017-04-03
Fix merging race conditions

### LINKED ISSUE #406: Undefined method each for Nil in result_merger

jenseng:
Ensure we cache appropriately so we don't merge more than once or do unsafe reads of the resultset.

See https://gist.github.com/jenseng/62465f674f8c02de09ef776f23d4dca4 for simple repro script.

Fixes #406 and may partially address #559 (though enforcing covered_percentage still requires workarounds à la [this comment](https://github.com/colszowka/simplecov/issues/559#issuecomment-280805651), since any process before the last one won't have all the coverage data)

Basically there are 3 main problems when using merging:

1. `SimpleCov.result` doesn't cache the `@result`, so the default `at_exit` behavior causes it to store and merge 3 times.
2. `SimpleCov::ResultMerger.resultset` calls `.stored_data` twice.
3. `SimpleCov::ResultMerger.merged_result` doesn't synchronize or use a cached `.resultset`, so a concurrent `.store_result` call can cause us to read an empty file.

This can cause the formatter to miss out on coverage data in our formatters and/or get the wrong values for covered percentages.

Furthermore, if you use OJ, `JSON.parse("") -> nil`, which means `.resultset` can be nil, causing exceptions as seen in #406. So I went ahead and rolled that fix into this, since `|| {}` is trivial, and the bug can easily be verified with my repro script above.

### DIFF
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -75,23 +75,21 @@ module SimpleCov
     # from cache using SimpleCov::ResultMerger if use_merging is activated (default)
     #
     def result
-      # Ensure the variable is defined to avoid ruby warnings
-      @result = nil unless defined?(@result)
+      return @result if result?
 
       # Collect our coverage result
-      if running && !result?
+      if running
         @result = SimpleCov::Result.new add_not_loaded_files(Coverage.result)
       end
 
       # If we're using merging of results, store the current result
-      # first, then merge the results and return those
+      # first (if there is one), then merge the results and return those
       if use_merging
         SimpleCov::ResultMerger.store_result(@result) if result?
-
-        SimpleCov::ResultMerger.merged_result
-      else
-        @result
+        @result = SimpleCov::ResultMerger.merged_result
       end
+
+      @result
     ensure
       self.running = false
     end
@@ -157,6 +155,13 @@ module SimpleCov
       rescue LoadError
         false
       end
+    end
+
+    #
+    # Clear out the previously cached .result. Primarily useful in testing
+    #
+    def clear_result
+      @result = nil
     end
   end
 end
--- a/lib/simplecov/result_merger.rb
+++ b/lib/simplecov/result_merger.rb
@@ -17,25 +17,31 @@ module SimpleCov
         File.join(SimpleCov.coverage_path, ".resultset.json.lock")
       end
 
-      # Loads the cached resultset from JSON and returns it as a Hash
+      # Loads the cached resultset from JSON and returns it as a Hash,
+      # caching it for subsequent accesses.
       def resultset
-        if stored_data
-          begin
-            JSON.parse(stored_data)
-          rescue
+        @resultset ||= begin
+          data = stored_data
+          if data
+            begin
+              JSON.parse(data) || {}
+            rescue
+              {}
+            end
+          else
             {}
           end
-        else
-          {}
         end
       end
 
       # Returns the contents of the resultset cache as a string or if the file is missing or empty nil
       def stored_data
-        return unless File.exist?(resultset_path)
-        data = File.read(resultset_path)
-        return if data.nil? || data.length < 2
-        data
+        synchronize_resultset do
+          return unless File.exist?(resultset_path)
+          data = File.read(resultset_path)
+          return if data.nil? || data.length < 2
+          data
+        end
       end
 
       # Gets the resultset hash and re-creates all included instances
@@ -76,8 +82,9 @@ module SimpleCov
 
       # Saves the given SimpleCov::Result in the resultset cache
       def store_result(result)
-        File.open(resultset_writelock, "w+") do |f|
-          f.flock(File::LOCK_EX)
+        synchronize_resultset do
+          # Ensure we have the latest, in case it was already cached
+          clear_resultset
           new_set = resultset
           command_name, data = result.to_hash.first
           new_set[command_name] = data
@@ -86,6 +93,28 @@ module SimpleCov
           end
         end
         true
+      end
+
+      # Ensure only one process is reading or writing the resultset at any
+      # given time
+      def synchronize_resultset
+        # make it reentrant
+        return yield if defined?(@resultset_locked) && @resultset_locked
+
+        begin
+          @resultset_locked = true
+          File.open(resultset_writelock, "w+") do |f|
+            f.flock(File::LOCK_EX)
+            yield
+          end
+        ensure
+          @resultset_locked = false
+        end
+      end
+
+      # Clear out the previously cached .resultset
+      def clear_resultset
+        @resultset = nil
       end
     end
   end
--- a/spec/result_merger_spec.rb
+++ b/spec/result_merger_spec.rb
@@ -1,10 +1,16 @@
 require "helper"
+require "tempfile"
+require "timeout"
 
 if SimpleCov.usable?
   describe SimpleCov::ResultMerger do
+    before do
+      SimpleCov::ResultMerger.clear_resultset
+      File.delete(SimpleCov::ResultMerger.resultset_path) if File.exist?(SimpleCov::ResultMerger.resultset_path)
+    end
+
     describe "with two faked coverage resultsets" do
       before do
-        SimpleCov.use_merging true
         @resultset1 = {
           source_fixture("sample.rb") => [nil, 1, 1, 1, nil, nil, 1, 1, nil, nil],
           source_fixture("app/models/user.rb") => [nil, 1, 1, 1, nil, nil, 1, 0, nil, nil],
@@ -73,15 +79,92 @@ if SimpleCov.usable?
               expect(SimpleCov::ResultMerger.results.length).to eq(1)
             end
           end
+        end
+      end
+    end
 
-          context "with merging disabled" do
-            before { SimpleCov.use_merging false }
+    describe ".store_result" do
+      it "refreshes the resultset" do
+        set = SimpleCov::ResultMerger.resultset
+        SimpleCov::ResultMerger.store_result({})
+        new_set = SimpleCov::ResultMerger.resultset
+        expect(new_set).not_to be(set)
+      end
 
-            it "returns nil for SimpleCov.result" do
-              expect(SimpleCov.result).to be_nil
+      it "persists to disk" do
+        SimpleCov::ResultMerger.store_result("a" => [1])
+        SimpleCov::ResultMerger.clear_resultset
+        new_set = SimpleCov::ResultMerger.resultset
+        expect(new_set).to eq("a" => [1])
+      end
+
+      it "synchronizes writes" do
+        expect(SimpleCov::ResultMerger).to receive(:synchronize_resultset)
+        SimpleCov::ResultMerger.store_result({})
+      end
+    end
+
+    describe ".resultset" do
+      it "caches" do
+        set = SimpleCov::ResultMerger.resultset
+        new_set = SimpleCov::ResultMerger.resultset
+        expect(new_set).to be(set)
+      end
+
+      it "synchronizes reads" do
+        expect(SimpleCov::ResultMerger).to receive(:synchronize_resultset)
+        SimpleCov::ResultMerger.resultset
+      end
+    end
+
+    describe ".synchronize_resultset" do
+      it "is reentrant (i.e. doesn't block its own process)" do
+        # without @resultset_locked, this spec would fail and
+        # `.store_result` wouldn't work
+        expect do
+          Timeout.timeout(1) do
+            SimpleCov::ResultMerger.synchronize_resultset do
+              SimpleCov::ResultMerger.synchronize_resultset do
+              end
             end
           end
+        end.not_to raise_error
+      end
+
+      it "blocks other processes" do
+        file = Tempfile.new("foo")
+
+        other_process = open("|ruby -e " + Shellwords.escape(<<-CODE) + " 2>/dev/null")
+          require "simplecov"
+          SimpleCov.coverage_dir(#{SimpleCov.coverage_dir.inspect})
+
+          # ensure the parent process has enough time to get a
+          # lock before we do
+          sleep 0.5
+
+          $stdout.sync = true
+          puts "running" # see `sleep`s in parent process
+
+          SimpleCov::ResultMerger.synchronize_resultset do
+            File.open(#{file.path.inspect}, "a") { |f| f.write("process 2\n") }
+          end
+        CODE
+
+        SimpleCov::ResultMerger.synchronize_resultset do
+          # wait until the child process is going, and then wait some more
+          # so we can be sure it blocks on the lock we already have.
+          sleep 0.1 until other_process.gets == "running\n"
+          sleep 1
+
+          # despite the sleeps, this will be written first since we got
+          # the first lock
+          File.open(file.path, "a") { |f| f.write("process 1\n") }
         end
+
+        # wait for it to finish
+        other_process.gets
+
+        expect(file.read).to eq("process 1\nprocess 2\n")
       end
     end
   end
--- a/spec/simplecov_spec.rb
+++ b/spec/simplecov_spec.rb
@@ -1,0 +1,109 @@
+require "helper"
+
+if SimpleCov.usable?
+  describe SimpleCov do
+    describe ".result" do
+      before do
+        SimpleCov.clear_result
+        allow(Coverage).to receive(:result).once.and_return({})
+      end
+
+      context "with merging disabled" do
+        before do
+          allow(SimpleCov).to receive(:use_merging).once.and_return(false)
+        end
+
+        context "when not running" do
+          before do
+            allow(SimpleCov).to receive(:running).and_return(false)
+          end
+
+          it "returns nil" do
+            expect(SimpleCov.result).to be_nil
+          end
+        end
+
+        context "when running" do
+          before do
+            allow(SimpleCov).to receive(:running).and_return(true, false)
+          end
+
+          it "uses the result from Coverage" do
+            expect(Coverage).to receive(:result).once.and_return(__FILE__ => [0, 1])
+            expect(SimpleCov.result.filenames).to eq [__FILE__]
+          end
+
+          it "adds not-loaded-files" do
+            expect(SimpleCov).to receive(:add_not_loaded_files).once.and_return({})
+            SimpleCov.result
+          end
+
+          it "doesn't store the current coverage" do
+            expect(SimpleCov::ResultMerger).not_to receive(:store_result)
+            SimpleCov.result
+          end
+
+          it "doesn't merge the result" do
+            expect(SimpleCov::ResultMerger).not_to receive(:merged_result)
+            SimpleCov.result
+          end
+
+          it "caches its result" do
+            result = SimpleCov.result
+            expect(SimpleCov.result).to be(result)
+          end
+        end
+      end
+
+      context "with merging enabled" do
+        let(:the_merged_result) { double }
+
+        before do
+          allow(SimpleCov).to receive(:use_merging).once.and_return(true)
+          allow(SimpleCov::ResultMerger).to receive(:store_result).once
+          allow(SimpleCov::ResultMerger).to receive(:merged_result).once.and_return(the_merged_result)
+        end
+
+        context "when not running" do
+          before do
+            allow(SimpleCov).to receive(:running).and_return(false)
+          end
+
+          it "merges the result" do
+            expect(SimpleCov.result).to be(the_merged_result)
+          end
+        end
+
+        context "when running" do
+          before do
+            allow(SimpleCov).to receive(:running).and_return(true, false)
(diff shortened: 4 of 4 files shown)


jenseng:
hah, looks like rubocop needs to be pinned even more precisely ... master fails on 0.48.1 but passes on 0.48.0 😭

PragTob:
Hi there!

Thanks a ton for this :tada: 

Might take me a little while to get around to a proper review, feel free to ping me if I haven't gotten to it within a week (goal is sooner, but you know... things).

I just noticed and fixed the rubocop thing... changes in patch levels :frowning: Master should be good after this so you can rebase/merge whatever.

Sorry for the inconvenience!

jenseng:
Rebased to get the rubocop fix 🤞

lib/simplecov.rb
```diff
@@ -75,23 +75,21 @@ def add_not_loaded_files(result)
     # from cache using SimpleCov::ResultMerger if use_merging is activated (default)
     #
     def result
-      # Ensure the variable is defined to avoid ruby warnings
```
jenseng: This was ostensibly added for rubocop, but rubocop and ruby don't complain about it now, so ¯\_(ツ)_/¯

spec/result_merger_spec.rb
```diff
@@ -73,14 +72,23 @@
...
+      describe ".store_result" do
+        it "refreshes the resultset" do
+          set = SimpleCov::ResultMerger.resultset
+          SimpleCov::ResultMerger.store_result({})
+          new_set = SimpleCov::ResultMerger.resultset
+          expect(new_set).not_to equal(set)
+        end
+      end
 
-            it "returns nil for SimpleCov.result" do
```
jenseng: This spec was misleading in that it only returned nil if `.running == false`, and it also happened independently of `use_merging` ... so it strikes me as not very useful, and a better spec can be found in spec/simplecov_spec.rb:22

Note that the behavior has now changed slightly; previously `.result` always returned `nil` if SimpleCov wasn't running. Now it only returns `nil` if it's not running **and** `use_merging == false`. If `use_merging == true`, you always get something useful from `result` (whether or not it merges the current `Coverage.result` depends on `.running`). This is useful for times when you have a post-CI script for formatting results and coverage thresholds, and the script itself doesn't need coverage (say, you have a script that aggregates and merges `.resultset.json` from many distributed workers, and want to format and enforce coverage on the final result).

jenseng:
It occurs to me it might be worth reworking the `flock`ing so that it **always** happens when you read. Like if you have a process reading the resultset without writing to it first (no `SimpleCov.start` call), it could still be affected by a concurrent writer and see an empty file.

e.g. could do something like

```ruby
def synchronize_resultset
  # make it reentrant
  return yield if @resultset_locked

  begin
    @resultset_locked = true
    File.open(resultset_writelock, "w+") do |f|
      f.flock(File::LOCK_EX)
      result = yield
    end
    result
  ensure
    @resultset_locked = false
  end
end

def stored_data
  synchronize_resultset do
    return unless File.exist?(resultset_path)
    data = File.read(resultset_path)
    return if data.nil? || data.length < 2
    data
  end
end

def store_result(result)
  synchronize_resultset do
    new_set = resultset(true)
    command_name, data = result.to_hash.first
    new_set[command_name] = data
    File.open(resultset_path, "w+") do |f|
      f.puts JSON.pretty_generate(new_set)
    end
  end
  true
end
```

That said, that would be a bigger change than the simple method caching I'm introducing in this pull request, so I'm happy to do that as a separate issue/PR.

bootstraponline:
would be awesome to have this merged 🌟

PragTob:
I agree. sorry I've been rather busy :|

I know I probably said this before but I'll try to make time this weekend to have a look :)

Sorry for the delay!

bootstraponline:
ping ⭐️

spec/result_merger_spec.rb
```diff
@@ -73,14 +72,23 @@
...
-              expect(SimpleCov.result).to be_nil
-            end
-          end
+      describe ".resultset" do
+        it "caches by default" do
+          set = SimpleCov::ResultMerger.resultset
+          new_set = SimpleCov::ResultMerger.resultset
+          expect(new_set).to equal(set)
```
PragTob: Not sure, but if we don't change the files etc. those should be the same whether we use merging or not, correct?

spec/simplecov_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+            SimpleCov.result
+          end
+
+          it "adds not-loaded-files" do
+            expect(SimpleCov).to receive(:add_not_loaded_files).once.and_return({})
+            SimpleCov.result
+          end
+
+          it "doesn't store the current coverage" do
+            expect(SimpleCov::ResultMerger).to receive(:store_result).never
```
PragTob: any reason against doing `not_to receive` ?

spec/simplecov_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+            SimpleCov.result
+          end
+
+          it "doesn't merge the result" do
+            expect(SimpleCov::ResultMerger).to receive(:merged_result).never
```
PragTob: same thing about `not_to` :)

spec/simplecov_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+        end
+
+        context "when running" do
+          before do
+            allow(SimpleCov).to receive(:running).and_return(true, false)
+          end
+
+          it "uses the result from Coverage" do
+            expect(Coverage).to receive(:result).once.and_return({})
+            SimpleCov.result
```
PragTob: somehow feels like it would be useful to see what `result` returns

spec/simplecov_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+          end
+
+          it "adds not-loaded-files" do
+            expect(SimpleCov).to receive(:add_not_loaded_files).once.and_return({})
```
PragTob: as this is basically ourselves, I'd think rather testing the effect that `add_not_loaded_files` has would make sense instead of mocking it. What do you think?

spec/simplecov_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+          end
+
+          it "doesn't merge the result" do
+            expect(SimpleCov::ResultMerger).to receive(:merged_result).never
+            SimpleCov.result
+          end
+
+          it "caches its result" do
+            result = SimpleCov.result
+            expect(SimpleCov.result).to equal(result)
```
PragTob: as long as we don't change any data the result should be equal itself with or without caching, or am I mistaken?

spec/simplecov_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+    describe ".result" do
+      before do
+        SimpleCov.clear_result!
+        allow(Coverage).to receive(:result).once.and_return({})
+        allow(SimpleCov).to receive(:add_not_loaded_files).once.and_return({})
+      end
+
+      context "with merging disabled" do
+        before do
+          allow(SimpleCov).to receive(:use_merging).once.and_return(false)
```
PragTob: couldn't we say `SimpleCov.use_merging(false)`?

spec/simplecov_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+            expect(SimpleCov.result).to equal(result)
+          end
+        end
+      end
+
+      context "with merging enabled" do
+        let(:the_merged_result) { double }
+
+        before do
+          allow(SimpleCov).to receive(:use_merging).once.and_return(true)
```
PragTob: couldn't we just do `SimpleCov.use_merging(false)` ?

spec/simplecov_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+          allow(SimpleCov::ResultMerger).to receive(:merged_result).once.and_return(the_merged_result)
+        end
+
+        context "when not running" do
+          before do
+            allow(SimpleCov).to receive(:running).and_return(false)
+          end
+
+          it "merges the result" do
+            expect(SimpleCov.result).to equal(the_merged_result)
```
PragTob: not quite sure but an empty double might be equal to any other empty double. Better give it a name or something like `double "some merged result"`

spec/simplecov_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+        end
+
+        context "when running" do
+          before do
+            allow(SimpleCov).to receive(:running).and_return(true, false)
+          end
+
+          it "uses the result from Coverage" do
+            expect(Coverage).to receive(:result).once.and_return({})
+            SimpleCov.result
```
PragTob: seems like we might wanna assert the return value?

spec/simplecov_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+          end
+
+          it "adds not-loaded-files" do
+            expect(SimpleCov).to receive(:add_not_loaded_files).once.and_return({})
```
PragTob: as add_not_loaded_files is on the thing under test I thinkw e mgiht opt to test what it does not that it gets called, wdyt?

lib/simplecov.rb
```diff
@@ -158,6 +156,13 @@ def usable?
         false
       end
     end
+
+    #
+    # Clear out the previously cached .result. Primarily useful in testing
+    #
+    def clear_result!
```
PragTob: Imo there should only be `!` versions of methods that exist without `!` to show slightly more dangerous ones so I'd opt to go without this :)

spec/simplecov_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+          it "merges the result" do
+            expect(SimpleCov.result).to equal(the_merged_result)
+          end
+
+          it "caches its result" do
+            result = SimpleCov.result
+            expect(SimpleCov.result).to equal(result)
+          end
+        end
+      end
```
PragTob: Thanks for speccing all this out!

PragTob:
The other idea seems good btw. - thank you. I'm just not sure how the `return yield if @resultset_locked` is supposed to work, as it just yields to the block aka executing it or am I missing something?

spec/result_merger_spec.rb
jenseng: the key point here is the rspec `equal` matcher calls `#equal?`, which checks if they are the same ruby object, not just if they are equivalent. with caching they are the same object, without they are not.

spec/simplecov_spec.rb
jenseng: 👍 sounds good, i'll change this

spec/simplecov_spec.rb
jenseng: i suppose, if we set it back. this feels safer IMO, since we're not changing instance variables of the SimpleCov class, which may have side effects in later specs

spec/simplecov_spec.rb
jenseng: nope, i can change it. either way sounds fine to me.

spec/simplecov_spec.rb
jenseng: see earlier `#equal?` note

spec/simplecov_spec.rb
jenseng: `#equal?` :)

spec/simplecov_spec.rb
jenseng: hmm, well i'd say `SimpleCov.result` is the thing under test, not `SimpleCov`, so i kind of like this test ¯\_(ツ)_/¯ ... maybe `add_not_loaded_files` could/should by handled by a separate object (though that's a bigger refactor) ? at any rate, it feels weird for a `.result` test to have to stub/fake a bunch of `File`/`Dir` stuff. though i do agree some sort of test around `add_not_loaded_files`' behavior would be nice

jenseng:
the idea is that `store_result` acquires a lock, and then calls `resultset` (source code elided in my example), which then calls `stored_data`, which should not actually need to acquire a lock since it already has one, so that's where `@resultset_locked` comes in

it's possible that `flock`ing in ruby is reentrant (i haven't checked), so that may not actually be necessary, which would simplify it a bit.

at any rate, i'll try to push up a tweaked PR in the next few days

spec/result_merger_spec.rb
PragTob: Personally I prefer `be` over `equal` for that because `equal` is so close to `eq` and `eql` that it's way too confusing that one checks equality, the other identity. I think `be` communicates that better, I can also change it myself though :)

spec/simplecov_spec.rb
PragTob: hm, fair point thanks! :+1:

spec/simplecov_spec.rb
PragTob: I prefer `not_to` but can also do the change myself if needbe. Thanks a lot :)

spec/simplecov_spec.rb
PragTob: Not wanting to stub a bunch of file stuff is a perfectly fine reason for stubbing early :+1: I somehow thought the effects would be easier to test, but with this I much prefer the current stub. True, maybe a refactoring for another object but I'd have to dig deeper to see what scope would make sense there.

PragTob:
@jenseng thank you a lot, let me know if you need some help or something so that we can get this released :tada:

spec/result_merger_spec.rb
jenseng: Oh hah, I had always assumed `be` did an `==` check, not `.equal?`... i'll change these `equal` matchers to `be` instead 👍

spec/simplecov_spec.rb
jenseng: It's a little weird here because `.result` just gives us the `the_merged_result` it got from `SimpleCov::ResultMerger` ... so I don't know how useful it is to check the return value here.

That said, I tweaked the similar spec on line 31 (non merging scenario) to show that the Coverage data does indeed get used

jenseng:
Ok, I believe it should be good to go... I think I addressed all the feedback, and commented where it didn't make sense to change it IMO. I also went ahead and made [this change](https://github.com/colszowka/simplecov/pull/570#issuecomment-291593459) so that reads are also synchronized.

Let me know if I missed anything, thanks!

lib/simplecov/result_merger.rb
```diff
@@ -87,6 +94,25 @@ def store_result(result)
...
         true
       end
+
+    private
+
+      # Ensure only one process is reading or writing the resultset at any
+      # given time
+      def synchronize_resultset
+        # make it reentrant
+        return yield if defined?(@resultset_locked) && @resultset_locked
```
PragTob: why do we need the `defined?` check? It should default to `nil`

PragTob: Also still not sure what making it reentrant here means and accomplishes, but reentrancy also has been a long time ago for me. The way I see it, the block we are given is now also called when the result set is already locked. Doesn't this effectively circumvent the lock?

PragTob: All of that said, do you happen to know of a way to test this? We could probably extract synchronize_resultset and yse that to write some specs around it which might be worthwhile, not sure. I just see code that I have questions about and often feel better, the better that code is tested :D

jenseng: It does, but it issues a ruby warning `instance variable @resultset_locked not initialized`, which fails the build, thanks to spec/support/fail_rspec_on_ruby_warning.rb

jenseng: Yes, so the idea is that if we call `synchronize_resultset` when we are already inside a `synchronize_resultset` call, we just let it through, since we know we already have a lock from the outer call. So we're not circumventing the lock, so much as saying "we already have a lock higher in the call stack, so just run the block". Due to how flock works, we would get a deadlock if we didn't, since the inner flock would wait forever for the outer one to release.

The reason we need to do this is we want to make both reads and writes acquire a lock so that neither one does anything unsafe. And in order to safely do a write, we 1. first need to read and 2. we need to ensure nobody else writes between our read and our write, otherwise we might lose data. So to accomplish that, write does `synchronize_resultset` around the whole thing so that any other processes have to wait. And since its nested read will also do `synchronize_resultset`, we use `@resultset_locked` to let it know it doesn't actually need a lock since we already got one.

jenseng: Sure, I have a couple ideas, I'll tweak the PR with some more tests

jenseng:
Ok, the latest tweak adds 5 new specs to spec/result_merger_spec.rb (4 around synchronization, 1 around persisting to disk), bringing the total new specs in this pull request up to 19 😅

lib/simplecov/result_merger.rb
PragTob: Which works as we are just one process, got it, thanks for taking the time to explain! :)

spec/result_merger_spec.rb
```diff
@@ -73,15 +79,81 @@
...
+        end.not_to raise_error
+      end
+
+      it "blocks other processes" do
+        file = Tempfile.new("foo")
+
+        fork do
+          # ensure the parent process has enough time to get a
+          # lock before we do
+          sleep 0.5
```
PragTob: Fork is not implemented on JRuby... :'( can we go and launch a separate one off script as a replacement?

With JRuby, we'd also need to wait way longer for that but that might be switched on whether we are running JRuby or not... or we could just switch the whole spec if it's too much of a hassle.

```
  1) SimpleCov::ResultMerger.synchronize_resultset blocks other processes

     Failure/Error: fork do

     

     NotImplementedError:

       fork is not available on this platform

     # ./spec/result_merger_spec.rb:137:in `block in (root)'
```

spec/result_merger_spec.rb
```diff
@@ -73,15 +79,81 @@
...
+        SimpleCov::ResultMerger.synchronize_resultset do
+          sleep 1
+          # despite the sleep, this will be written first since we got
+          # the first lock
+          File.open(file.path, "a") { |f| f.write("process 1\n") }
         end
+
+        Process.wait
+
+        expect(file.read).to eq("process 1\nprocess 2\n")
```
PragTob: :clap:

jenseng:
Ok hopefully that should do it 🤞 ... spec works more or less the same as before, except now it uses `open` to run some ruby commands, rather than `fork`ing the current process

PragTob:
Yup, thank you a lot. Time for some rest after all that hard work! :sleeping: 

img_20170713_145446

jenseng:
Woohoo, thanks! 🎉 🍰 🍺

PR #571, opened 2017-04-03
Rubocop fix

PragTob:
Yay patch level breaks :)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,7 @@ group :test do
     gem "cucumber"
     gem "phantomjs", "~> 2.1"
     gem "poltergeist"
-    gem "rubocop", "~> 0.48.0" unless RUBY_VERSION.start_with?("1.")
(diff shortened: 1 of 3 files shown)


jenseng:
With this code change, you might consider also pinning `~> 0.48.1`, otherwise it now fails if you have `0.48.0` in your bundle. Repro:

1. check out this PR
2. `rm Gemfile.lock` (if you have one)
3. `gem uninstall rubocop -v 0.48.1` (if you have it)
4. `gem install rubocop -v 0.48.0` (if you don't already have it)
5. `bundle install`
6. `rake` -> 💥

jenseng:
Or alternatively it looks like you could set `Style/PercentLiteralDelimiters` so as to be explicit ... see https://github.com/bbatsov/rubocop/issues/4039

PragTob:
it.... what? Thanks for the info I'll just lock down to the patch level now...

PR #578, opened 2017-05-06
switch inject to sum

abrahamchaibi:
In certain cases I get an error for inject(&:+) however things work fine for inject(:+) and sum. This change to use inject(:+) seems to solve it for me across ruby versions whereas sum fails on older rubies. Otherwise have to use a monkeypatched version which is a bit of a shame :(

### DIFF
--- a/lib/simplecov/file_list.rb
+++ b/lib/simplecov/file_list.rb
@@ -5,25 +5,25 @@ module SimpleCov
     # Returns the count of lines that have coverage
     def covered_lines
       return 0.0 if empty?
-      map { |f| f.covered_lines.count }.inject(&:+)
+      map { |f| f.covered_lines.count }.inject(:+)
     end
 
     # Returns the count of lines that have been missed
     def missed_lines
       return 0.0 if empty?
-      map { |f| f.missed_lines.count }.inject(&:+)
+      map { |f| f.missed_lines.count }.inject(:+)
     end
 
     # Returns the count of lines that are not relevant for coverage
     def never_lines
       return 0.0 if empty?
-      map { |f| f.never_lines.count }.inject(&:+)
+      map { |f| f.never_lines.count }.inject(:+)
     end
 
     # Returns the count of skipped lines
     def skipped_lines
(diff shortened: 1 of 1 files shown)


abrahamchaibi:
Is there a reason you chose to use inject(&:+) overall?

abrahamchaibi:
for example:
`irb(main):004:0* [1,2,3].inject(:+)
=> 6
irb(main):005:0> [1,2,3].sum
=> 6
irb(main):006:0> [1,2,3].inject(&:+)
`

PragTob:
Hi there!

Thanks for the PR!

Which ruby version are you using there? Our tests run against 1.9+ and some basic stuff against 1.8 so it'd be odd for the tests not to catch. And I think Symbol#to_proc was added in 1.9? So it seems odd to cause failures.

The change should be good regardless, just wondering myself.

abrahamchaibi:
Thanks for the quick response! I am on Ruby 2.3.0 using active_support core_ext. Based on the definition of sum [here](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/enumerable.rb#L31) I think they even define sum using `inject(:+)` due to an issue with ruby 2.4 omitted identity in inject. Not sure why this would not be caught by the tests in ruby 2.3.3 though. Might be that it only comes up with active support?

PragTob:
```
tobi@speedy ~ $ ruby -v
ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-linux]
tobi@speedy ~ $ irb
irb(main):002:0> [1,2,3].inject(&:+)
=> 6
irb(main):003:0> require 'active_support'
=> true
irb(main):004:0> require 'active_support/core_ext'
=> true
irb(main):005:0> [1,2,3].inject(&:+)
=> 6
```

Doesn't seem to be, that said I also installed the newest active_support... anyhow it's just an interest thing, the change should be good and most likely even faster either way.

Thanks!

abrahamchaibi:
thank you!!

PR #583, opened 2017-05-18
Merge duplicate files

matrinox:
- It checks similarity based on file content
- Combines coverage of files with same content (we could do it by lowercase name however that assumes the computer is using a case-insensitive file system)
- Chooses filename of the file that exists (close enough)

### DIFF
--- a/lib/simplecov/result.rb
+++ b/lib/simplecov/result.rb
@@ -27,7 +27,14 @@ module SimpleCov
       @original_result = original_result.freeze
       @files = SimpleCov::FileList.new(original_result.map do |filename, coverage|
         SimpleCov::SourceFile.new(filename, coverage) if File.file?(filename)
-      end.compact.sort_by(&:filename))
+      end.compact.sort_by(&:filename)).group_by do |file|
+        File.read(file.filename).hash
+      end.map do |_, files|
+        files.reduce(nil) do |acc, file|
+          next file unless acc
+          combine_files(acc, file)
+        end
+      end
       filter!
     end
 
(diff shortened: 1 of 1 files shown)


matrinox:
> However, I'm not so sure I like this feature. People having duplicate files in their project feels more like a problem I wouldn't wanna solve in simplecov. People should just have one file and deduplicate it themselves (and I really hope this doesn't happen all that often).

Sorry, I should've explained myself better. It solves duplicate require paths. As in:
```ruby
require 'lib/foo.rb'
require 'lib/Foo.rb'
```

In a case-insensitive filesystem, like the default on Macs, these will always be the same file (if it exists). Your feedback suggests to me that you think that people should just rename it but I feel like it's a small improvement to help typos (a small case typo can split the stats).

> For the future, it's often better to first open an issue and discuss a possible feature to see if and in what form the feature is wanted, to avoid implementing something that might not make it in.

I forked it for myself and "patched" this because I wanted it for myself (and I like programming for fun). But thanks for the suggestion!

PragTob:
Hi there,

thanks for the clarification. It's weird to me that the standard library `coverage` works with require paths and not the real file name :thinking: (or maybe we do that, would need to check)

It's hard to solve then, really. As you say, just lowercasing it would assume case insensitive file system and might break stuff. I wonder if there's another way to check if 2 paths resolve to the same file?

I assume people should fix it themselves, is the easiest solution in the end.  It's not nice to say, but I can't come up with a better way (except if there was another way to check that it resolves to the same path). The current solution works, but it introduces a non negligible performance overhead (hashing, reading files, etc) that users would have to pay every time they run tests/CI (it's important to our users that simplecov has a small overhead). I should benchmark it to be certain, but I'm pretty sure with file reads, hashing and system commands.

Speaking of system commands, the current one relies on them which would break Windows, sure we could introduce separate windows commands but that'd also be quite the overhead (conceptually) for us to maintain.

Still thanks a lot for the contribution, and if you have an idea for another solution, please let me know!

matrinox:
> thanks for the clarification. It's weird to me that the standard library coverage works with require paths and not the real file name 🤔 (or maybe we do that, would need to check)

@PragTob My guess is that when the file is loaded, the system passes the path that was used to find it.

> It's hard to solve then, really. As you say, just lowercasing it would assume case insensitive file system and might break stuff. I wonder if there's another way to check if 2 paths resolve to the same file?

I check the file contents. Not the most perfect method but pretty reliable. Hard to imagine someone has the same file contents with the same file name but different case. Still, it's not perfect. Here's a [smarter example](https://stackoverflow.com/questions/7870041/check-if-file-system-is-case-insensitive-in-python) in Python (I'm sure a Ruby translation is possible)

> The current solution works, but it introduces a non negligible performance overhead (hashing, reading files, etc) that users would have to pay every time they run tests/CI (it's important to our users that simplecov has a small overhead). I should benchmark it to be certain, but I'm pretty sure with file reads, hashing and system commands.

That's the first thing that popped into my head. I think a solution like creating a temp file and testing if the os can find one all upper case is a decent solution that should be much faster, since you only need to run it once. Once it is known that the file system is case-insensitive, we can convert all file paths to lowercase. That should be quite performant. I am willing to work on it this week as I have some time.

PragTob:
@matrinox thanks for your willingness to work on this. I think I'd rather not have the feature in one way or another. People shouldn't duplicate their files with the exact same content. Even if we did the name checking, if it includes the path it's futile either way. And files might as well be called the same in the same path. If people do this and it annoys them they can just delete one file and can be all the better happier. Still thanks for the contribution!

matrinox:
@PragTob. I don't mean multiple files with the same content, I mean one file with different file paths. Something like this...
```ruby
# File 1
require 'lib/foo.rb'
# File 2
require 'lib/Foo.rb'
```

They could go an rename the requires in every single file that has this but it's not quite the same as deleting the other file.

PragTob:
ah right, sorry. Yes, but still that's more of an easy to fix "they" problem. Especially as eventually someone wants to deploy to/run on  linux were this should fail and break either way.

Issue #419, opened 2015-10-01
Groups use absolute paths


MaienM:
It seems that the groups match the full path, instead of just the part relative to the directory it's all being run from. I noticed this when on my jenkins server, all files where included in the Libraries group (with the filter `lib`), because the files were located somewhere deep in `/var/lib/jenkins/`.

Now this may very well be intentional, but it feels very illogical to me.

This is using simplecov v0.10.0, ruby 2.2.3, rails-api 4.2.4, and rspec 3.3.2.

bf4:
What's `SimpleCov.root` ?  where are you configuring it? How are you running it? Where do you 'start' SimpleCov?  What's an example of an expected path that is included and an unexpected one? Thanks

MaienM:
Sorry for the somewhat delayed response.

`Simplecov.root` is 

```
/var/lib/jenkins/jobs/AEB/workspace
```

I am configuring it through a `.simplecov` file, contents are as follows

``` ruby
require 'simplecov-rcov'
SimpleCov.formatters = [
	SimpleCov::Formatter::HTMLFormatter,
	SimpleCov::Formatter::RcovFormatter,
]
SimpleCov.profiles.define 'aeb' do
	@filters = []

	add_filter '/spec/'
	add_filter '/config/'

	add_group 'Controllers' do |src_file|
		src_file.filename['app/controllers'] && !src_file.filename['concern']
	end
	add_group 'Concerns', 'app/controllers/concerns'
	add_group 'Models', 'app/models'
	add_group 'Docs', 'app/docs'
	add_group 'Libraries' do |src_file|
		src_file.filename['lib/middleware'] || src_file.filename['lib/other']
	end
end
SimpleCov.start 'aeb'
```

As you can see, the Libraries group now checks for certain subfolders of lib manually, which is the workaround we're currently using. That line used to be

``` ruby
add_group 'Libraries', 'lib'
```

I'm running it with the RSpec test cases, by placing `require 'simplecov'` as the first thing in `spec/spec_helper.rb`. Rspec is run by executing `bin/bundle exec rake spec`. 

The working directory while running these commands is

```
/var/lib/jenkins/jobs/AEB/workspace
```

Examples of expected paths are:

```
/var/lib/jenkins/jobs/AEB/workspace/lib/other/permission_initializer.rb
/var/lib/jenkins/jobs/AEB/workspace/lib/ext/string.rb
/var/lib/jenkins/jobs/AEB/workspace/lib/ci/run.rb
[...]

PR #589, opened 2017-06-16
Add regex filters

### LINKED ISSUE #514: Default filters for Rails profile exclude unexpected files
Currently, default profile for `:rails` adds some filters, for example `/config/` and `/db/` filters.

Because of how filters are implemented this will also exclude files located in `/lib/module/my/config/for/something/` which is a little unexpected.

It would be great to make it possible to improve files to match against relative routes like `\A/config/` or something like that, what do you think?

Thanks for the great work with SimpleCov!

jsteel:
Add the ability to use regex filters. The regexes should match against the files path relative to the current project, not the full path to root. For example, the file /Users/jon/project/app/foo/test.rb would match on a filter of "jon". I made the String and Array matchers follow that new rule as well.

Fixes #514

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,7 +262,7 @@ report.
 
 ### Defining custom filters
 
-You can currently define a filter using either a String (that will then be Regexp-matched against each source file's path),
+You can currently define a filter using either a String or Regexp (that will then be Regexp-matched against each source file's path),
 a block or by passing in your own Filter class.
 
 #### String filter
@@ -274,6 +274,16 @@ end
 ```
 
 This simple string filter will remove all files that match "/test/" in their path.
+
+#### Regex filter
+
+```ruby
+SimpleCov.start do
+  add_filter %r{^/test/}
+end
+```
+
+This simple regex filter will remove all files that start with /test/ in their path.
 
 #### Block filter
 
@@ -304,6 +314,17 @@ SimpleCov.add_filter LineFilter.new(5)
 Defining your own filters is pretty easy: Just inherit from SimpleCov::Filter and define a method 'matches?(source_file)'. When running
 the filter, a true return value from this method will result in the removal of the given source_file. The filter_argument method
 is being set in the SimpleCov::Filter initialize method and thus is set to 5 in this example.
+
+#### Array filter
+
+```ruby
+SimpleCov.start do
+  proc = Proc.new { |source_file| fales }
+  add_filter ["string", /regex/, proc, LineFilter.new(5)]
+end
+```
+
+You can pass in an array containing any of the other filter types.
 
 #### Ignoring/skipping code
 
--- a/doc/editor-integration.md
+++ b/doc/editor-integration.md
@@ -7,6 +7,11 @@ Some editors have a graphical integration for the simplecov gem.
 
 Adds an overview of your current test coverage to Atom.
 
+#### [Sublime Editor: Simple​Cov](https://packagecontrol.io/packages/SimpleCov)
+*by sentience*
+
+Adds in editor live coverage highlighting, status bar coverage information, and summary coverage information.
+
 #### [cadre](https://github.com/nyarly/cadre)
 *by Judson Lester*
 
--- a/lib/simplecov/configuration.rb
+++ b/lib/simplecov/configuration.rb
@@ -293,12 +293,10 @@ module SimpleCov
     def parse_filter(filter_argument = nil, &filter_proc)
       if filter_argument.is_a?(SimpleCov::Filter)
         filter_argument
-      elsif filter_argument.is_a?(String)
-        SimpleCov::StringFilter.new(filter_argument)
       elsif filter_proc
         SimpleCov::BlockFilter.new(filter_proc)
-      elsif filter_argument.is_a?(Array)
-        SimpleCov::ArrayFilter.new(filter_argument)
+      elsif filter_argument
+        SimpleCov::Filter.class_for_argument(filter_argument).new(filter_argument)
       else
         raise ArgumentError, "Please specify either a string or a block to filter with"
       end
--- a/lib/simplecov/configuration.rb
+++ b/lib/simplecov/configuration.rb
@@ -293,10 +293,8 @@ module SimpleCov
     def parse_filter(filter_argument = nil, &filter_proc)
       filter = filter_argument || filter_proc
 
-      if filter_argument.is_a?(SimpleCov::Filter)
-        filter_argument
-      elsif filter
-        SimpleCov::Filter.class_for_argument(filter).new(filter)
+      if filter
+        SimpleCov::Filter.build_filter(filter)
       else
         raise ArgumentError, "Please specify either a filter or a block to filter with"
       end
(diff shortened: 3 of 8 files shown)


lib/simplecov/configuration.rb
```diff
@@ -293,12 +293,10 @@ def add_group(group_name, filter_argument = nil, &filter_proc)
...
       if filter_argument.is_a?(SimpleCov::Filter)
         filter_argument
-      elsif filter_argument.is_a?(String)
-        SimpleCov::StringFilter.new(filter_argument)
       elsif filter_proc
         SimpleCov::BlockFilter.new(filter_proc)
-      elsif filter_argument.is_a?(Array)
-        SimpleCov::ArrayFilter.new(filter_argument)
+      elsif filter_argument
+        SimpleCov::Filter.class_for_argument(filter_argument).new(filter_argument)
```
PragTob: Cool stuff! :+1:

jsteel:
No problem. I'll give the README some TLC today.

README.md
```diff
@@ -305,6 +315,17 @@ Defining your own filters is pretty easy: Just inherit from SimpleCov::Filter an
...
+#### Array filter
+
+```ruby
+SimpleCov.start do
+  proc = Proc.new { |source_file| fales }
+  add_filter ["string", /regex/, proc, LineFilter.new(5)]
+end
+```
+
+You can pass in an array containing any of the other filter types.
```
PragTob: Thanks! :+1:

lib/simplecov/filter.rb
```diff
@@ -63,11 +65,23 @@ def matches?(source_file)
   end
 
   class ArrayFilter < SimpleCov::Filter
-    # Returns true if any of the file paths passed in the given array matches the string
-    # configured when initializing this Filter with StringFilter.new(['some/path', 'other/path'])
+    def initialize(filter_argument)
+      filter_argument.map! do |arg|
```
PragTob: I think we should rather not use `map!` as I don't wanna mutate whatever was passed in, so rather without `!` :)

lib/simplecov/filter.rb
```diff
@@ -63,11 +65,23 @@ def matches?(source_file)
...
+        if arg.is_a?(SimpleCov::Filter)
+          arg
+        else
+          Filter.class_for_argument(arg).new(arg)
```
PragTob: just thinking out loud, couldn't we also put that check into `Filter.class_for_argument` e.g. if it is already a filter? Would make that code easier and centralize the knowledge :)

lib/simplecov/filter.rb
```diff
@@ -63,11 +65,23 @@ def matches?(source_file)
...
+        end
+      end
+
+      super(filter_argument)
+    end
```
PragTob: I have a feeling, the additions here were necessary because THINGS weren't working. Sounds like we might miss some specs here. If you wanna add those, great! :+1: 

If not let me know and I'll try to get to it when I got some time :)

lib/simplecov/filter.rb
PragTob: Ah wait we can't because this one already instantiates which we wouldn't need there... maybe another helper method? Not quite sure.

lib/simplecov/filter.rb
jsteel: Good call.

lib/simplecov/filter.rb
jsteel: Yah sounds good. I'll make a factory function that instantiates based the appropriate class type, or does nothing if it's given a filter.

lib/simplecov/filter.rb
jsteel: There is definitely a hole in the specs. I had to do manual tests to make sure this worked. I'll see if I can whip up some specs.

lib/simplecov/filter.rb
```diff
@@ -25,7 +25,14 @@ def passes?(source_file)
       matches?(source_file)
     end
 
+    def self.build_filter(filter_argument)
+      return filter_argument if filter_argument.is_a?(SimpleCov::Filter)
+      class_for_argument(filter_argument).new(filter_argument)
+    end
```
PragTob: :heart_eyes:

doc/editor-integration.md
```diff
@@ -7,7 +7,12 @@ Some editors have a graphical integration for the simplecov gem.
 
 Adds an overview of your current test coverage to Atom.
 
+#### [Sublime Editor: Simple​Cov](https://packagecontrol.io/packages/SimpleCov)
+*by sentience*
+
+Adds in editor live coverage highlighting, status bar coverage information, and summary coverage information.
```
PragTob: :+1: drive by fixes

lib/simplecov/filter.rb
```diff
@@ -25,7 +25,14 @@ def passes?(source_file)
...
+
     def self.class_for_argument(filter_argument)
+      return filter_argument if filter_argument.is_a?(SimpleCov::Filter)
```
PragTob: do we need this? I thought if needed this should be handled by `.build_filter` ?

lib/simplecov/filter.rb
```diff
@@ -66,15 +73,15 @@ def matches?(source_file)
 
   class ArrayFilter < SimpleCov::Filter
     def initialize(filter_argument)
-      filter_argument.map! do |arg|
+      filter_objects = filter_argument.map do |arg|
         if arg.is_a?(SimpleCov::Filter)
           arg
         else
           Filter.class_for_argument(arg).new(arg)
```
PragTob: didn't we wanna use `.build_filter` here? :)

lib/simplecov/filter.rb
jsteel: Definitely not. Not sure how I put that in there.

lib/simplecov/filter.rb
jsteel: Right thanks. I knew there was a reason I abstracted that out.

jsteel:
This is a more correct way to get the projects directory, agreed?

jsteel:
The changes in here will also fix #419 I believe

lib/simplecov/source_file.rb
```diff
@@ -82,8 +82,7 @@ def initialize(filename, coverage)
 
     # The path to this source file relative to the projects directory
     def project_filename
-      project_dir = File.dirname(File.expand_path(File.basename(__FILE__)))
-      @filename.sub(/^#{project_dir}/, "")
+      @filename.sub(/^#{SimpleCov.root}/, "")
```
PragTob: hm not too sure but I think it looks good :+1: 

In the future though, would be preferred to have unrelated changes for other issues etc. in their own PR so they don't get too big and stay more focussed.

jsteel: Agreed, but this change was kind of required for the regex. Without it the regex would be matching the full path, which probably nobody would understand.

PragTob: Ah I see, didn't make that connection in my head. Thanks!

PR #594, opened 2017-06-28
Fix small "fales" typo in array filter section of README

dhedlund:
The README contains a typo in code example, where `fales` should be `false`.


PragTob:
Thank you!

PR #614, opened 2017-08-25
Add another SimpleCov formatter

ysksn:
This formatter is used to emulate t_wada AA (not exact though).

- t_wada is Japanese famous TDD practitioner
- actual AA is like [this](http://twitpic.com/dzgwk8)

thanks :)


PragTob:
Thank you :green_heart:

PR #616, opened 2017-08-27
Stop handling string filters as regexps

yujinakayama:
Previously, filter strings passed to `add_filter` have been handled as regexps. It means, for example, `add_filter '.pl'` matches against `sample.rb` because `.` is handled as _any character_ in regexp.

As now we have [regexp filter](#589), there's no reason to handle filter strings as regexps.

However we need to think about semver with this change. If this behavior was intentional, we cannot introduce this change until next major version bump. Or we can do if this was a _bug_.

What do you think?

### DIFF
--- a/spec/filters_spec.rb
+++ b/spec/filters_spec.rb
@@ -31,10 +31,6 @@ if SimpleCov.usable?
       expect(SimpleCov::StringFilter.new(parent_dir_name)).not_to be_matches subject
     end
 
-    it "matches a new SimpleCov::StringFilter '/fixtures/'" do
-      expect(SimpleCov::StringFilter.new("sample.rb")).to be_matches subject
-    end
-
     it "matches a new SimpleCov::RegexFilter /\/fixtures\//" do
       expect(SimpleCov::RegexFilter.new(/\/fixtures\//)).to be_matches subject
     end
--- a/spec/filters_spec.rb
+++ b/spec/filters_spec.rb
@@ -26,6 +26,10 @@ if SimpleCov.usable?
       expect(SimpleCov::StringFilter.new("sample.rb")).to be_matches subject
     end
 
+    it "doesn't match a new SimpleCov::StringFilter '.pl'" do
+      expect(SimpleCov::StringFilter.new(".pl")).not_to be_matches subject
+    end
+
     it "doesn't match a parent directory with a new SimpleCov::StringFilter" do
       parent_dir_name = File.basename(File.expand_path("..", File.dirname(__FILE__)))
       expect(SimpleCov::StringFilter.new(parent_dir_name)).not_to be_matches subject
--- a/lib/simplecov/filter.rb
+++ b/lib/simplecov/filter.rb
@@ -49,7 +49,7 @@ module SimpleCov
     # Returns true when the given source file's filename matches the
     # string configured when initializing this Filter with StringFilter.new('somestring)
     def matches?(source_file)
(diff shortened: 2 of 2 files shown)


PragTob:
:wave: 

Hi, first thanks for your contribution :tada: 

This is an interesting issue and a very astute observation! A problem is that for all that I can tell [it has been behaving like this forever](https://github.com/colszowka/simplecov/blame/5d1103816f83dba8257d024fd5a94ef7c53226a5/lib/simple_cov.rb).

However, the [README doesn't seem to mention that it has regexp like behaviour](https://github.com/colszowka/simplecov#string-filter):

All in all, I'm in favor, the example you provided is especially frightening :scream: So thanks for that!

There's another issue I might need to reach out to Christoph for, just wanna double check with him as I'm not aware of all the history/intentions.

Regarding version, in `0.y.z` versions you don't need a major version bump for changes:

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

If we fix it I'd internally classify it as a "bug" but as it breaks behaviour old as time I'd still bump a minor to release it so people can have a proper warning looking at the changelog because patch level updates (imo even in 0.y.z) should be safe.

spec/filters_spec.rb
```diff
@@ -26,15 +26,15 @@
...
+    end
+
     it "doesn't match a parent directory with a new SimpleCov::StringFilter" do
       parent_dir_name = File.basename(File.expand_path("..", File.dirname(__FILE__)))
       expect(SimpleCov::StringFilter.new(parent_dir_name)).not_to be_matches subject
     end
 
-    it "matches a new SimpleCov::StringFilter '/fixtures/'" do
-      expect(SimpleCov::StringFilter.new("sample.rb")).to be_matches subject
-    end
```
PragTob: why id you remove this spec, shouldn't that still work?

yujinakayama: Because these are exactly same (though the latter description is wrong):

https://github.com/colszowka/simplecov/blob/c5e8654834e922d3e69a980aef029d4bcc880b6e/spec/filters_spec.rb#L25-L27
https://github.com/colszowka/simplecov/blob/c5e8654834e922d3e69a980aef029d4bcc880b6e/spec/filters_spec.rb#L34-L36

Maybe a mistake of copy-pasting?

yujinakayama:
> If we fix it I'd internally classify it as a "bug" but as it breaks behaviour old as time I'd still bump a minor to release it so people can have a proper warning looking at the changelog because patch level updates (imo even in 0.y.z) should be safe.

I forgot the fact that simplecov is still under 1.0 👍

spec/filters_spec.rb
PragTob: Ah yes, missed that! Seems like it was meant to be `"/fixtures/"` indeed! Unless I missed that one as well :)

colszowka:
Hey all, sorry for my late reply! Thanks for the PR, looks very reasonable to me!

As to the potentially breaking change:

Personally, I'd say I originally built it like this out of shear laziness. I agree that it should not automatically expand into a regexp under the hood and magically. I also agree that this could be a breaking change for existing users, since a significant behaviour changes. Under SemVer that should warrant a major version bump. We really should finally declare something 1.0 :grin: I think what @PragTob suggested, to bump minor version should be fine :rocket:

PragTob:
Well then :shipit: 

Thanks @colszowka

PR #617, opened 2017-08-29
Filter directories outside SimpleCov.root that have it as a prefix

jenseng:
Given project directory `/foo/app` and BUNDLE_PATH `/foo/app_cache`,
ensure the latter is filtered out correctly by the default root filter,
since it's outside of SimpleCov.root.

Otherwise you'll see all those gem files in your coverage report, with
nonsensical paths like `._cache/ruby/2.4.0/gems...`, unless you add a
nonsensical `/^_cache/` filter

Also add some test coverage for the default filters

### DIFF
--- a/lib/simplecov/defaults.rb
+++ b/lib/simplecov/defaults.rb
@@ -6,7 +6,7 @@ SimpleCov.profiles.define "root_filter" do
   # Exclude all files outside of simplecov root
   root_filter = nil
   add_filter do |src|
-    root_filter ||= /\A#{Regexp.escape(SimpleCov.root)}/io
+    root_filter ||= /\A#{Regexp.escape(SimpleCov.root + File::SEPARATOR)}/io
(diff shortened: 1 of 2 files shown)


jenseng:
Paths are hard, I fixed essentially the same bug in bundler a while back 😂  ... https://github.com/bundler/bundler/pull/4715

lib/simplecov/defaults.rb
```diff
@@ -6,7 +6,7 @@
   # Exclude all files outside of simplecov root
   root_filter = nil
   add_filter do |src|
-    root_filter ||= /\A#{Regexp.escape(SimpleCov.root)}/io
+    root_filter ||= /\A#{Regexp.escape(SimpleCov.root + File::SEPARATOR)}/io
```
PragTob: :man_facepalming:

PR #618, opened 2017-08-29
Fix regression in rails filters

jenseng:
In converting this to a regex in a0b14a1, an extra leading `/` was added,
making this filter less useful (it would literally match the `/^/db/`
path).

### DIFF
--- a/spec/defaults_spec.rb
+++ b/spec/defaults_spec.rb
@@ -1,0 +1,41 @@
+require "helper"
+
+if SimpleCov.usable?
+  describe SimpleCov do
+    skip "requires the default configuration" if ENV["SIMPLECOV_NO_DEFAULTS"]
+
+    context "profiles" do
+      let(:config_class) do
+        Class.new do
+          include SimpleCov::Configuration
+
+          def load_profile(name)
+            configure(&SimpleCov.profiles[name.to_sym])
+          end
+        end
+      end
+
+      let(:config) { config_class.new }
+
+      def filtered?(config, filename)
+        path = File.join(SimpleCov.root, filename)
+        file = SimpleCov::SourceFile.new(path, [nil, 1, 1, 1, nil, nil, 1, 0, nil, nil])
+        config.filters.any? { |filter| filter.matches?(file) }
+      end
+
+      it "provides a sensible test_frameworks profile" do
+        config.load_profile(:test_frameworks)
+        expect(filtered?(config, "foo.rb")).not_to be
+        expect(filtered?(config, "test/foo.rb")).to be
+        expect(filtered?(config, "spec/bar.rb")).to be
+      end
+
+      it "provides a sensible rails profile" do
+        config.load_profile(:rails)
+        expect(filtered?(config, "app/models/user.rb")).not_to be
+        expect(filtered?(config, "db/schema.rb")).to be
+        expect(filtered?(config, "config/environment.rb")).to be
+      end
+    end
+  end
+end
--- a/spec/filters_spec.rb
+++ b/spec/filters_spec.rb
@@ -155,7 +155,9 @@ if SimpleCov.usable?
       end
     end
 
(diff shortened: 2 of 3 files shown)


PragTob:
:wave: 

Thanks! Could we add a "simple" test case that checks this maybe in the integration test cases so something like this doesn't happen again? :)

jenseng:
Sure thing 👍

jenseng:
This approach to testing is a little bit meh... ideally `.filtered` and friends could be moved onto an object or into `Configuration` so that they are more easily testable without worrying about breaking the singleton. That feels like a bigger refactor than this little fix though, so ¯\\_(ツ)_/¯

jenseng:
Oh, heh I think I glossed over your comment around feature tests, perhaps I'll just test it there

jenseng:
Upon further reflection, I think I found a decent way to test it at a lower level 🤞

spec/filters_spec.rb
```diff
@@ -155,7 +155,9 @@ def matches?(_)
       end
     end
 
-    context "with the default profile" do
+    context "with the default configuration" do
+      skip "requires the default configuration" if ENV["SIMPLECOV_NO_DEFAULTS"]
```
PragTob: Slightly confused why this is here and not just above? Leftover? :)

spec/defaults_spec.rb
```diff
@@ -0,0 +1,41 @@
+require "helper"
+
+if SimpleCov.usable?
+  describe SimpleCov do
+    skip "requires the default configuration" if ENV["SIMPLECOV_NO_DEFAULTS"]
```
PragTob: (with other point where it is I mean this here) :)

spec/defaults_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    context "profiles" do
+      let(:config_class) do
+        Class.new do
+          include SimpleCov::Configuration
+
+          def load_profile(name)
+            configure(&SimpleCov.profiles[name.to_sym])
+          end
+        end
+      end
```
PragTob: :+1: using let with anonymous classes! :)

spec/defaults_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+        config.load_profile(:test_frameworks)
+        expect(filtered?(config, "foo.rb")).not_to be
+        expect(filtered?(config, "test/foo.rb")).to be
+        expect(filtered?(config, "spec/bar.rb")).to be
+      end
+
+      it "provides a sensible rails profile" do
+        config.load_profile(:rails)
+        expect(filtered?(config, "app/models/user.rb")).not_to be
+        expect(filtered?(config, "db/schema.rb")).to be
```
PragTob: wait whaaaaatttttttttt? You can just use `.to be`/`.not_to be`. Touché. I've used rspec forever and consider myself pretty good at it bu you constantly teach me new stuff. Thanks :pray:

spec/defaults_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+        expect(filtered?(config, "config/environment.rb")).to be
```
PragTob: :+1: great way of testing it on a lower level, thank you very much! :+1:

spec/filters_spec.rb
jenseng: I'm not sure if there's a scenario where you'd run the specs w/ SIMPLECOV_NO_DEFAULTS, but this should ensure the suite still passes in that case, since it has to do with the default filters. Might be worth setting up in Travis?

spec/defaults_spec.rb
jenseng: Same story here, I just wanted to make this resilient in case people run specs with that set, since the profiles won't exist in that case.

PragTob:
Thanks a bunch! 🚀

PR #623, opened 2017-09-11
Enforce frozen string literal pragma comments

PragTob:


PragTob:
as I feared, auto added the lines also in the example projects etc. influencing the line counts :woman_facepalming:

PR #625, opened 2017-10-01
Fixes #624 Checks max_coverage_drop on failed tests

### LINKED ISSUE #624: Failing tests can cause coverage drop when refusing coverage drop
Currently simplecov only checks ```maximum_coverage_drop``` on a successful test run, so a failing test run can write to ```.last_run.json``` with a lower coverage percenage than before.

Here is a scenario:

test coverage is at 100%
configure simplecov to refuse coverage drop, or set max coverage drop to 0
run specs
```.last_run.json``` shows 100%
write some new code
write a spec that only partially covers the new code that fails
run specs
```.last_run.json``` now shows < 100%
fix spec to pass, but test coverage still less that 100%
run specs
```.last_run.json``` still shows < 100%

I have PR that I will raise shortly that fixes this

### PR DESCRIPTION
**Pull request by thomas07vt**
Fixes #624 

Just taking out the max coverage drop check outside the success if statement. I can't think of a reason to leave it in the success if statement, but of course I am not too familiar with this codebase :) . Can you think of a reason why we would want to write to ```.last_run.json``` on failing tests?

The only thing that I wasn't sure of is, if a test is failing, maybe we want to leave the pre-existing non-zero exit status rather than writing an exit status of 3.

### DIFF
--- a/lib/simplecov/defaults.rb
+++ b/lib/simplecov/defaults.rb
@@ -85,11 +85,10 @@ at_exit do # rubocop:disable Metrics/BlockLength
           @exit_status = SimpleCov::ExitCodes::MAXIMUM_COVERAGE_DROP
         end
       end
-    end
 
-    # Don't overwrite last_run file if refuse_coverage_drop option is enabled and the coverage has dropped
-    unless @exit_status == SimpleCov::ExitCodes::MAXIMUM_COVERAGE_DROP
-      SimpleCov::LastRun.write(:result => {:covered_percent => covered_percent})
+      if @exit_status == SimpleCov::ExitCodes::SUCCESS # rubocop:disable Metrics/BlockNesting
+        SimpleCov::LastRun.write(:result => {:covered_percent => covered_percent})
+      end
     end
   end
 
--- a/features/maximum_coverage_drop.feature
+++ b/features/maximum_coverage_drop.feature
@@ -85,5 +85,55 @@ Feature:
         }
       }
       """
+  Scenario: test failures do not update the resultset
+    Given SimpleCov for RSpec is configured with:
+      """
+      require 'simplecov'
+      SimpleCov.start do
+        add_group 'Libs', 'lib/faked_project/'
+        add_filter '/spec/'
+        maximum_coverage_drop 0
+      end
+      """
 
+    And a file named "lib/faked_project/missed.rb" with:
+      """
+      class UncoveredSourceCode
+        def foo
+          never_reached
+        rescue => err
+          but no one cares about invalid ruby here
+        end
+      end
+      """
+
+    And a file named "spec/failing_spec.rb" with:
+      """
+      require "spec_helper"
+      describe FakedProject do
+        it "fails" do
+          expect(false).to eq(true)
+        end
+      end
+      """
+    And the file named "coverage/.last_run.json" with:
+      """
+      {
+        "result": {
+          "covered_percent": 100.0
+        }
+      }
+      """
+
+    When I run `bundle exec rspec spec`
+    Then the exit status should be 1
+    And a file named "coverage/.last_run.json" should exist
+    And the file "coverage/.last_run.json" should contain:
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by PragTob** [comment]
Hi, first thanks for your contribution! I'll take a look, but first it seems that the build is failing because rubocop inspects some vendored gems... grml.

**Comment by thomas07vt** [comment]
Yeah I checked the ```.rubocop.yml``` file and its using an exclude path that seems like it should work ```'vendor/bundle/**/*'``` so I am not sure what's going on...

**Comment by PragTob** [comment]
and that failure is a bug in rubocop... PRing then workaround fixing. Then reviewing here hopefully :)

**Comment by thomas07vt** [comment]
Oh, you have already done that!... closing this one :)

**Comment by PragTob** [comment]
Nope, this one we sitll need

**Comment by thomas07vt** [comment]
> I understand the bug correctly this should also solve it, right? :)

yeah that should work too. I can update my PR to add that change.

**Comment by PragTob** [comment]
That'd be great!

**Comment by thomas07vt** [comment]
@PragTob would you prefer the duplicate ```write``` call as I have it in the PR, or a duplicate check on ```@exit_status`` like this:

```ruby
    if @exit_status == SimpleCov::ExitCodes::SUCCESS # No other errors
      if covered_percent < SimpleCov.minimum_coverage # rubocop:disable Metrics/BlockNesting
        $stderr.printf("Coverage (%.2f%%) is below the expected minimum coverage (%.2f%%).\n", covered_percent, SimpleCov.minimum_coverage)
        @exit_status = SimpleCov::ExitCodes::MINIMUM_COVERAGE
      elsif covered_percentages.any? { |p| p < SimpleCov.minimum_coverage_by_file } # rubocop:disable Metrics/BlockNesting
        $stderr.printf("File (%s) is only (%.2f%%) covered. This is below the expected minimum coverage per file of (%.2f%%).\n", SimpleCov.result.least_covered_file, covered_percentages.min, SimpleCov.minimum_coverage_by_file)
        @exit_status = SimpleCov::ExitCodes::MINIMUM_COVERAGE
      elsif (last_run = SimpleCov::LastRun.read) # rubocop:disable Metrics/BlockNesting
        coverage_diff = last_run["result"]["covered_percent"] - covered_percent
        if coverage_diff > SimpleCov.maximum_coverage_drop # rubocop:disable Metrics/BlockNesting
          $stderr.printf("Coverage has dropped by %.2f%% since the last time (maximum allowed: %.2f%%).\n", coverage_diff, SimpleCov.maximum_coverage_drop)
          @exit_status = SimpleCov::ExitCodes::MAXIMUM_COVERAGE_DROP
        end
      end
    end

    if @exit_status == SimpleCov::ExitCodes::SUCCESS
      SimpleCov::LastRun.write(:result => {:covered_percent => covered_percent})
    end
```

I kinda like the duplicate ```@exit_status``` check better, but I'll defer to you :)

lib/simplecov/defaults.rb
```diff
@@ -83,14 +83,13 @@
         if coverage_diff > SimpleCov.maximum_coverage_drop # rubocop:disable Metrics/BlockNesting
           $stderr.printf("Coverage has dropped by %.2f%% since the last time (maximum allowed: %.2f%%).\n", coverage_diff, SimpleCov.maximum_coverage_drop)
           @exit_status = SimpleCov::ExitCodes::MAXIMUM_COVERAGE_DROP
+        else
+          SimpleCov::LastRun.write(:result => {:covered_percent => covered_percent})
         end
+      else
+        SimpleCov::LastRun.write(:result => {:covered_percent => covered_percent})
```
PragTob: I think I'd just move the write tot he bottom of the bigger if and keep the `exit_status` check as before I think. But this whole code is ripe for some proper refactoring though :)

thomas07vt:
@PragTob How about this?

PragTob:
Should be good enough, thanks! :)

PR #631, opened 2017-10-07
Break default profiles out into new files

thomas07vt:
The goal behind this PR is to simplify the ```defaults.rb``` file. It
just breaks out the default profiles into their own files to
reduce the noise in this ```defaults.rb``` file

Things I like about this PR:

- It simplifies ```defaults.rb```
- It gives structure to any more default profiles that will be added, or
other default files to be added.

Things I don't like about this PR:

- Typically folders are used for namespacing and the ```defaults/``` folder
is not used in that way in the PR.
- There will probably be very few if any more default profiles added, so
point 2 above has little weight.
- The files inside```defaults/``` actually sets up default profiles, so
maybe we should use a ```profiles/``` folder? or ```default_profiles/```
folder? Although a ```default/``` folder is generic enough to add say
a ```config.rb``` file that has the ```SimpleCov.configure do``` setup.

What do you think? Worth the refactor?

### DIFF
--- a/lib/simplecov/defaults.rb
+++ b/lib/simplecov/defaults.rb
@@ -3,43 +3,10 @@
 # Load default formatter gem
 require "simplecov-html"
 require "pathname"
-
-SimpleCov.profiles.define "root_filter" do
-  # Exclude all files outside of simplecov root
-  root_filter = nil
-  add_filter do |src|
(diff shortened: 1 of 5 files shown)


thomas07vt:
👍 Ok I renamed the folder to ```profiles```. Let me know if you want me to make anymore changes.

Issue #439, opened 2015-12-07
Simplecov can't generate reports of files with names with non-ASCII characters


guapolo:
I'm using Simplecov 0.11.1, ruby 2.2.3.

I have a file with a non-ASCII character (ñ). When running Simplecov, I get the following error:

```
(erb):18:in `concat': incompatible character encodings: UTF-8 and ASCII-8BIT (Encoding::CompatibilityError)
    from (erb):18:in `block in formatted_source_file'
    from (erb):14:in `each'
    from (erb):14:in `each_with_index'
    from (erb):14:in `formatted_source_file'
    from /Users/pruiz/.rvm/rubies/ruby-2.2.3/lib/ruby/2.2.0/erb.rb:863:in `eval'
    from /Users/pruiz/.rvm/rubies/ruby-2.2.3/lib/ruby/2.2.0/erb.rb:863:in `result'
    from /Users/pruiz/.rvm/gems/ruby-2.2.3@escolares/gems/simplecov-html-0.10.0/lib/simplecov-html.rb:56:in `formatted_source_file'
    from (erb):36:in `block (2 levels) in format'
    from (erb):35:in `each'
    from (erb):35:in `block in format'
    from /Users/pruiz/.rvm/rubies/ruby-2.2.3/lib/ruby/2.2.0/erb.rb:863:in `eval'
    from /Users/pruiz/.rvm/rubies/ruby-2.2.3/lib/ruby/2.2.0/erb.rb:863:in `result'
    from /Users/pruiz/.rvm/gems/ruby-2.2.3@escolares/gems/simplecov-html-0.10.0/lib/simplecov-html.rb:23:in `block in format'
    from /Users/pruiz/.rvm/gems/ruby-2.2.3@escolares/gems/simplecov-html-0.10.0/lib/simplecov-html.rb:22:in `open'
    from /Users/pruiz/.rvm/gems/ruby-2.2.3@escolares/gems/simplecov-html-0.10.0/lib/simplecov-html.rb:22:in `format'
    from /Users/pruiz/.rvm/gems/ruby-2.2.3@escolares/gems/simplecov-0.11.1/lib/simplecov/result.rb:47:in `format!'
    from /Users/pruiz/.rvm/gems/ruby-2.2.3@escolares/gems/simplecov-0.11.1/lib/simplecov/configuration.rb:168:in `block in at_exit'
[...]

PR #633, opened 2017-10-17
Remove unused fixtures

mistydemeo:
The tests associated with these fixtures look like they disappeared during the test-unit to rspec migration, but the fixtures themselves stayed around anyway.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,5 @@
 AllCops:
   Exclude:
-    - 'spec/fixtures/iso-8859.rb'
     - 'tmp/**/*'
     - 'vendor/bundle/**/*'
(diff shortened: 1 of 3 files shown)


PragTob:
:wave: 

Hi there! Thanks for the contribution and thanks for the digging :tada: 

As you say they look like they disappeared during the migration I can't help but think that this probably was an oversight. Especially since we seem to have encoding related bugs somewhere (see #439 ) do you think it'd be possible with appropriate effort to reinstate those specs using the fixtures? :)

Thanks! :green_heart:

PragTob:
I think we still want this and no activity --> need to figure out the tests that were missing as a recent issue #655 also brought up that one of these might be broken :|

PR #637, opened 2017-10-27
RFC: Ignore coverage on skipped lines

ereslibre:
The json resultset will include 0 passes on lines that are marked as
:nocov:. This can lead external tools to think that this line was not
in any pass, when it was effectively marked as to be ignored by the
coverage.

This PR is mainly a discussion on how we could achieve something
similar. I have three possible options (off the top of my head):

1) Discard this completely, this makes no sense on `simplecov`.
2) This PR: the speed is penalized in the general case, what is not
good at all. Plus, it modifies the `resultset.json` in a way that is not
what ruby would have written for coverage.
3) *Possibly my ideal solution*: we can allow an option to be provided
when we call to `SimpleCov.start` in which we can provide a different
file for the `:nocov:` sections on all files. This file would be again a hash
of filenames with arrays, with a boolean, marking if the line was in a
`:nocov:` section or not. If this option is not provided, everything runs
as default and this nocov file is not generated, so really fast as usual.

The rationale behind this change is that external tools processing the
`resultset.json` file have no indication about ignored lines, what makes
them hard to do calculations on them properly.

More background: https://github.com/codeclimate/test-reporter/issues/246

### DIFF
--- a/lib/simplecov/result.rb
+++ b/lib/simplecov/result.rb
@@ -12,6 +12,8 @@ module SimpleCov
     extend Forwardable
     # Returns the original Coverage.result used for this instance of SimpleCov::Result
     attr_reader :original_result
+    # Returns the final Coverage.result for this instance after skipped lines have been processed
+    attr_reader :final_result
     # Returns all files that are applicable to this result (sans filters!) as instances of SimpleCov::SourceFile. Aliased as :source_files
     attr_reader :files
     alias source_files files
@@ -27,8 +29,9 @@ module SimpleCov
     # coverage data)
     def initialize(original_result)
       @original_result = original_result.freeze
+      @final_result = Hash[@original_result.keys.zip([nil] * @original_result.keys.size)]
       @files = SimpleCov::FileList.new(original_result.map do |filename, coverage|
-        SimpleCov::SourceFile.new(filename, coverage) if File.file?(filename)
+        SimpleCov::SourceFile.new(filename, coverage, @final_result) if File.file?(filename)
       end.compact.sort_by(&:filename))
       filter!
     end
@@ -77,7 +80,7 @@ module SimpleCov
 
     def coverage
       keys = original_result.keys & filenames
-      Hash[keys.zip(original_result.values_at(*keys))]
+      Hash[keys.zip(final_result.values_at(*keys))]
     end
 
     # Applies all configured SimpleCov filters on this result's source files
--- a/lib/simplecov/source_file.rb
+++ b/lib/simplecov/source_file.rb
@@ -77,9 +77,11 @@ module SimpleCov
     # The array of coverage data received from the Coverage.result
     attr_reader :coverage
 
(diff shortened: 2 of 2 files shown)


ereslibre:
This is a brief diff of the two resultsets for this section exactly: https://github.com/smartbox-io/cell/blob/615856a31e36ba02ad54d6a5975441732da0c43d/app/controllers/api/v1/objects_controller.rb#L32-L36

```
~ diff resultset-no-patch.json resultset-with-patch.json 
242,246c242,246
<         null,
<         1,
<         0,
<         null,
<         null,
---
>         -1,
>         -1,
>         -1,
>         -1,
>         -1,
358c358
<     "timestamp": 1509119702
---
>     "timestamp": 1509127414
```

PragTob:
Hi there,

thanks for the report!

I'd need to do some further investigation into what getting `null` or whatever value is desired there returned into the resultset to say how feasible this is from our side.

However, my overall standpoint on this is that the `resultset.json`  isn't part of the official API. It's an implementation detail of how we merge parallel testsuites etc. It's a cache. For handling simplecov results there are [formatters](https://github.com/colszowka/simplecov#using-your-own-formatter). I don't know if this was every communicated in any other way. I also seem to remember codeclimate being implemented with a formatter before. I'm sure they had their reasons to change it, I just don't know of them. Imo the `resultset.json` isn't the best place to integrate - contents, location, name - all that _could_ change in future versions. It's not even documented. 

Cheers,
Tobi

ale7714:
@ereslibre @PragTob 👋 

I have a question about how in overall those a formatter knows that some lines are between a `:nocov:` block. For those cases they don't really on the cache (`.resultset.json`)? 

Would a possibility be to add a new key on `.resultset.json` called `"NoCoverage"` and include the number lines where the `:nocov:` block start/end per file?

Thank you for the feedback on this, I appreciate it.

PragTob:
@ale7714 :wave: 

Imo it doesn't look at the gathered coverage at all for those, we just mark all the lines in the code as skipped and are done with it.

I can just strongly advise, not to use `resultset.json` as a way to integrate - for instance there may be [branch coverage up coming up in ruby 2.5](https://bugs.ruby-lang.org/issues/13901). We might or might not change the format based on that so that we could have a unified format both for line and branch coverage. I haven't really looked into it much (time is sparse, sorry - hence the late reply).

Imo the best integration point would still be through a formatter - even if it just dumps a file in a format you control which is then processed by your go program. I've seen that you had some troubles you were using internal APIs for - we could look at those and either make them public or make them "non issues" by fixing something (I believe it was about merging the results). If the formatter API is lacking we can also figure that out, but the cache file as an integration point isn't really future proof imo.

Cheers,
Tobi

PragTob:
I'm gonna close this. `resultset.json` is not an official API. We have a very simple API for formatters that gives access to our whole result object - skipped lines and everything. That's the way integration was envisioned to work.

PR #639, opened 2017-11-01
Fixes #493 Capture exit exception for reference

### LINKED ISSUE #493: let me know what exception happened in at_exit
I'm doing some custom reporting and want to not do that when the user has Interrupted the test run,
can we pass in $! to the at_exit call ... or store it somewhere globally ?

thomas07vt:
Fixes #493

This stores any exit exception that may have occurred. The idea is to
expose this exception to the user supplied ```at_exit``` block.

```ruby
SimpleCov.at_exit do
  if SimpleCov.exit_exception
    # Do something or nothing due to an exception
  else
    # Do something or nothing due to no exception
  end
end
```

Another method was added: ```exit_status_from_exception``` to reduce the
logic in the at_exit block. I am not sure if it's worth adding this
method though. There is still opportunity to refactor the ```at_exit```
block. I first attempted to refactor the @exit_status logic more but
it will be a bigger change. If you want I can remove the
```exit_status_from_exception``` method from this PR

I also don't love having ```begin rescue end``` blocks in tests, but I wasn't sure
if there was a better way to test that. Maybe I could wrap ```$!/$ERROR_INFO```
in another method and mock that in testing.

### DIFF
--- a/spec/simplecov_spec.rb
+++ b/spec/simplecov_spec.rb
@@ -107,5 +107,68 @@ if SimpleCov.usable?
         end
       end
     end
+
+    describe ".set_exit_exception" do
+      context "when an exception has occurred" do
+        let(:error) { StandardError.new "SomeError" }
+
+        after do
+          # Clear the exit_exception
+          SimpleCov.set_exit_exception
+        end
+
+        it "captures the current exception" do
+          begin
+            raise error
+          rescue
+            SimpleCov.set_exit_exception
+            expect(SimpleCov.exit_exception).to be(error)
+          end
+        end
+      end
+
+      context "when an exception has not occurred" do
+        it "has no exit_exception" do
+          SimpleCov.set_exit_exception
+          expect(SimpleCov.exit_exception).to eq(nil)
+        end
+      end
+    end
+
+    describe ".exit_status_from_exception" do
+      context "when no exception has occurred" do
+        before do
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: spec/simplecov_spec.rb [hunk 1]
```diff
@@ -107,5 +107,89 @@
...
+            raise error
+          rescue
+            SimpleCov.set_exit_exception
+            expect(SimpleCov.exit_exception).to be(error)
+          end
+        end
+      end
+
+      context "when an exception has not occurred" do
+        it "does set the exit_exception" do
```
PragTob: `has no exit_exception` might be better here :)

spec/simplecov_spec.rb
```diff
@@ -107,5 +107,89 @@
...
+        end
+      end
+
+      after do
+        # Clear the exit_exception
+        SimpleCov.set_exit_exception
+      end
+
+      it "returns the exit exception" do
+        expect(SimpleCov.exit_exception).to be(error)
```
PragTob: imo we already tested this basically in the tests before this, so I'd be good without this spec :)

thomas07vt:
Thanks for the feedback. I'll implement the fixes in the next couple days 👍

thomas07vt:
and by next couple of days... i apparently meant the next 10 days.... :) OK, ready for another review.

PR #646, opened 2018-01-12
Allow docile 1.2

jaredbeck:
docile's first release in 3.5 years!

There do not appear to be any breaking changes, according
to https://github.com/ms-ati/docile/blob/master/HISTORY.md

### DIFF
--- a/simplecov.gemspec
+++ b/simplecov.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency "json", ">= 1.8", "< 3"
   gem.add_dependency "simplecov-html", "~> 0.10.0"
-  gem.add_dependency "docile", "~> 1.1.0"
+  gem.add_dependency "docile", "~> 1.1"
 
   gem.add_development_dependency "bundler", "~> 1.9"
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ unreleased
 
 ## Enhancements
 
+* Relax version constraint on `docile`, per SemVer
 * exception that occurred on exit is available as `exit_exception`! See [#639](https://github.com/colszowka/simplecov/pull/639)  (thanks @thomas07vt)
 
 ## Bugfixes


ms-ati:
Thanks @jaredbeck. Can confirm no breaking changes, just the addition of a new method. Following semantic versioning, this resulted in a minor version bump :)

ms-ati:
@colszowka Appears that the test failures are completely unrelated, can anyone help resolve the cucumber failures in rake so this can be merged?

ms-ati:
@PragTob I've pushed a PR that fixes the cucumber tests in #647

PragTob:
:wave: 

as the fix is on master I'll try rerunning the tests, if this works all great otherwise we might need to merge master or rebase.

ms-ati:
@jaredbeck would you be able to now rebase this PR onto the latest master and force push it?

simplecov.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency "json", ">= 1.8", "< 3"
   gem.add_dependency "simplecov-html", "~> 0.10.0"
-  gem.add_dependency "docile", "~> 1.1.0"
+  gem.add_dependency "docile", [">= 1.1.0", "< 1.3"]
```
ms-ati: @jaredbeck @PragTob Is this dependency declaration still unnecessarily restrictive?

Respecting semantic versioning, shouldn't it be sufficient to say:

```ruby
gem.add_dependency "docile", "~> 1.1"
```

Since the dependency practices semantic versioning, shouldn't upper-bound version be all the way to the indication of a breaking change, which would be `2.0`? So therefore it should be enough to say `"~> 1.1"`, which is equivalent to `">= 1.1.0", "< 2.0"`, right?

For reference: http://guides.rubygems.org/patterns/#declaring-dependencies

PragTob: yup it is. I just wasn't sure if docile followed semver (couldn't see it in the README) - so this was fine to me. If you say it does, then it's fine by me :) With relatively common dependencies such as simplecov I tend to be more conservative because more than a couple of times have new gem releases broken something for me.

jaredbeck:
>> So therefore it should be enough to say "~> 1.1" ..

> yup it is. I just wasn't sure if docile followed semver (couldn't see it in the README) - so this was fine to me. If you say it does, then it's fine by me :)

Changed constraint to `"~> 1.1"` and rebased. Thanks.

PragTob:
Thanks!

jaredbeck:
Is there anything I can do to help get this released? Thanks!

PragTob:
There's a couple of outstanding PRs, after that I'll probably cut another release. For the time being it can be used from git :)

jaredbeck:
This was released in 0.16.0 on 2018-03-15

PR #647, opened 2018-01-12
Pin bundler to 1.16.0

ms-ati:
This is a workaround for https://github.com/rubygems/rubygems/issues/2055
which seems to have worked for other projects tested on Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,8 @@ language: ruby
 
 before_install:
   - gem update --system
-  - gem install bundler
+  # Pin bundler version due to https://github.com/rubygems/rubygems/issues/2055
+  - gem install bundler -v 1.16.0
 
 bundler_args: --without development --jobs=3 --retry=3
 


ms-ati:
@PragTob This should fix the cucumber failures that were observed in https://github.com/colszowka/simplecov/pull/646 :)

ms-ati:
@colszowka this should fix your build issues, see rubygems/rubygems#2055

PragTob:
:wave:

Thanks!

The issue also seems to mention that a new rubygems versions fixes it so maybe a `rubygems update --system` or whatever could also fix it and might be better because we wouldn't be pinned to an older version :)

Wanna try that out?

ms-ati:
Hi @PragTob, please note that SimpleCov was already doing the `rubygems update --system`, and this PR doesn't change that.

Regardless of that particular statement on the issue, the observable fact is that the version of Rubygems that you get on Travis when you run that command is *not compatible* with Bundler 1.16.1.

Please note that the workaround of pinning Bundler to 1.16.0 in this PR is the workaround applied by many other projects -- you can see this at the  bottom of the linked issue, where the commits that reference the issue appear.

PragTob:
Hi there,

sorry forgot that we were already updating rubygems. Sometimes hard to keep track off all the projects :)

That a workaround is applied elsewhere is not a good enough point imo - workarounds work but there might be other underlying problems. See #649 that fixes this problem by removing the travis cache which seems to be a likely cause of this problem. Whehter the new bundler version just exposed it, or both in combination just cause it I can't tell.

Closed in favor of #649

ms-ati:
Ah cool. As far as I know you are the first person to discover this alternate workaround (deleting the bundler cache), so many thanks and kudos to you! Hope others can follow the trail to discover it as well.

PR #648, opened 2018-01-13
Ms ati fix cucumber in travis

PragTob:
Builds on #647 trying out force install mentioned in rubygems/rubygems#2055

also updates some rubies.


PragTob:
Closing here to keep the history for bug hunting on parts of rubygems, bundler and travis. Real fix is over in #649

PR #651, opened 2018-01-29
gemspec: Avoid distributing cucumber.yml

olleolleolle:
This PR excludes the `cucumber.yml` configuration file from the gem.

See https://github.com/cucumber/cucumber-rails/issues/346#issuecomment-361167227


PragTob:
Thanks a bunch!

PR #656, opened 2018-02-23
Update alternate-formatters.md with simplecov-erb

kpaulisse:
I needed to generate output from simplecov that displayed in the output of my CI jobs. I implemented this via a user-configurable ERB template, so that the output format can be customized with minimal friction. Source is at https://github.com/kpaulisse/simplecov-erb and it's available under the MIT license.


PragTob:
Sounds great, thank you!

PR #660, opened 2018-03-09
Update the CONTRIBUTING doc.

BMorearty:
The doc said simplecov uses Test::Unit but it uses RSpec and Cucumber.
Updated the doc to be correct, and fixed the examples of how to run
tests.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,8 +43,9 @@ If you want to contribute, please:
 
 ## Running Individual Tests
 
-This project uses Test::Unit. Individual tests can be run like this:
+This project uses RSpec and Cucumber. Individual tests can be run like this:
 
 ```bash
-ruby -I test path/to/test.rb
+bundle exec rspec path/to/test.rb
+bundle exec cucumber path/to/test.feature
 ```


CONTRIBUTING.md
```diff
@@ -31,7 +31,8 @@ To fetch & test the library for development, do:
     $ git clone https://github.com/colszowka/simplecov.git
     $ cd simplecov
     $ bundle
-    $ bundle exec rake
+    $ bundle exec rake spec
+    $ bundle exec cucumber
```
PragTob: Why was this changed? `rake` should just execute both of those!

BMorearty: TL;DR there was something wrong on my system. Please ignore this part of the PR and leave it as `bundle exec rake`.

When I ran `bundle exec rake` in Ruby 2.1.8, all that ran was Rubocop. And it flagged a bunch of errors. It did not run rspec or cucumber. I just realized why: since Rubucop flagged those errors, it returned a nonzero result. That caused Rake to stop.

But something is weird. When I run Rubocop by itself with `bundle exec rubocop` instead of `bundle exec rake`, it does not show errors. Stranger still: after that, `bundle exec rake` no longer shows Rubocop errors, so it proceeds to run `rspec` and `cucumber`.

I think this is related to a `rubocop-airbnb` gem that's on my system but not in the bundle. It has style rules that don't apply to simplecov. Why did it get loaded? I have no idea. Anyway, it works now. 😕

BMorearty: I will remove it...

PR #661, opened 2018-03-09
Make tests pass.

BMorearty:
- Ruby 1.9.3 could not bundle install because it complained about frozen
  strings RUBY_VERSION and SimpleCov::VERSION. Calling `dup` returns
  an unfrozen string.
- Tests weren't passing in any Ruby version because `require "stringio"`
  was happening too late. This resulted in:

```
Failure/Error: @stderr_stream = StringIO.new

NameError:
  uninitialized constant FailOnWarnings::StringIO
./spec/support/fail_rspec_on_ruby_warning.rb:9:in `initialize'
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ source "https://rubygems.org"
 # Uncomment this to use development version of html formatter from github
 # gem 'simplecov-html', :github => 'colszowka/simplecov-html'
 
-gem "rake", Gem::Version.new(RUBY_VERSION) < Gem::Version.new("1.9.3") ? "~>10.3" : ">= 10.3"
+gem "rake", Gem::Version.new(RUBY_VERSION.dup) < Gem::Version.new("1.9.3") ? "~>10.3" : ">= 10.3"
 
 group :test do
   gem "rspec", ">= 3.2"
--- a/simplecov.gemspec
+++ b/simplecov.gemspec
@@ -5,7 +5,7 @@ require "simplecov/version"
 
 Gem::Specification.new do |gem|
   gem.name        = "simplecov"
-  gem.version     = SimpleCov::VERSION
+  gem.version     = SimpleCov::VERSION.dup
   gem.platform    = Gem::Platform::RUBY
   gem.authors     = ["Christoph Olszowka"]
   gem.email       = ["christoph at olszowka de"]
--- a/spec/helper.rb
+++ b/spec/helper.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "rspec"
+require "stringio"
 # loaded before simplecov to also capture parse time warnings
 require "support/fail_rspec_on_ruby_warning"
 require "simplecov"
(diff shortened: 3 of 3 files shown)


PragTob:
Hi!

Thanks!

What's your setup - this seems weird. We run against all those ruby versions in CI and they (flakies aside) [pass just fine](https://travis-ci.org/colszowka/simplecov/builds/350339735)

BMorearty:
Hi @PragTob!

Regarding the `dup` calls: I noticed that your CI was working and was wondering why my results were different. The `Gem::Version` issue seems to be related to the version of Rubygems. My Ruby 1.9.3 is on an Ubuntu system ruby Rubygems 1.8.23. 

```
brian_morearty@i-029fc5432483fe35f:~/repos/simplecov$ gem -v
1.8.23
brian_morearty@i-029fc5432483fe35f:~/repos/simplecov$ irb
irb(main):001:0> ver = '1.9'.freeze
=> "1.9"
irb(main):002:0> Gem::Version.new(ver)
RuntimeError: can't modify frozen String
	from /usr/lib/ruby/1.9.1/rubygems/version.rb:191:in `strip!'
	from /usr/lib/ruby/1.9.1/rubygems/version.rb:191:in `initialize'
	from (irb):2:in `new'
	from (irb):2
	from /mnt/airlab/rbenv/versions/1.9.3/bin/irb:12:in `<main>'
```

So I tried `gem update --system` and I get Ubuntu's error message:

```
ERROR:  While executing gem ... (RuntimeError)
    gem update --system is disabled on Debian, because it will overwrite the content of the rubygems Debian package, and might break your Debian system in subtle ways. The Debian-supported way to update rubygems is through apt-get, using Debian official repositories.
If you really know what you are doing, you can still update rubygems by setting the REALLY_GEM_UPDATE_SYSTEM environment variable, but please remember that this is completely unsupported by Debian.
```

But I'm using rbenv so now I have rubygems/rbenv and Ubuntu competing with each other so it doesn't make sense to update rubygems with apt-get. 

I'll just leave my system alone and if you don't want to take the `dup` calls, I'm fine with that.

----

Regarding the `stringio` require statement in helper.rb, I don't know why your CI is working. By code inspection, it looks like the `stringio` require statement has to happen before requiring `support/fail_rspec_on_ruby_warning`. Here's why:

Without this change:

1. helper.rb line 5: `require "support/fail_rspec_on_ruby_warning"` 
1. fail_rspec_on_ruby_warning.rb line 7: defines `class FailOnWarnings`
1. fail_rspec_on_ruby_warning.rb line 70: `warning_collector = FailOnWarnings.new`
1. fail_rspec_on_ruby_warning.rb line 9: `FailOnWarnings` initializer calls `@stderr_stream = StringIO.new`. << This needs stringio.

PragTob:
Sounds good, thanks!

Seeing that this rubygems version is ancient and probably also riddled with security holes I wouldn't wanna take it in :)

The stringio fix I'll gladly take, thanks! (might be that something else is already loading it... would still be weird in the difference of environments but you're right!

PragTob:
Adjusted by hand and merged, thanks!

PR #663, opened 2018-03-09
Static asset server setting has nothing to do

vochicong:
In my Rails 5 app, 
`config.public_file_server.enabled = false` 
setting will make all UI tests (RSpec System spec) fail.

I believe that static asset server setting has nothing to do with Spring and SimpleCov.
`config.eager_load = false` is just enough.

For the same reason, for Rails 4 we may not need to set
`config.serve_static_files = false`
but I haven't checked it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -633,14 +633,8 @@ If you're using [Spring](https://github.com/rails/spring) to speed up test suite
 1. Change the following settings in `test.rb`.
 
     ```ruby
-    # For Rails 4 or earlier, use the following configuration:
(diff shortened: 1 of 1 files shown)


PragTob:
Hi!

Thanks! I think I don't use these settings either :angel: 

I could track this back to 263ee149759cd0ccb4 which doesn't seem to have a PR to it and no further explanation other than that nginx and Apache do it which you know normally I don't run development in so yeah - seems this can go. 

Thanks!

PR #664, opened 2018-03-10
Don't crash on invalid UTF-8 byte sequences.

BMorearty:
If a line of code has an invalid byte sequence in UTF-8, count it as a relevant line rather than crashing.

Please don't dismiss this as a useless thing to do. It turns out the [Builder gem has a line of code with an invalid UTF-8 byte sequence](https://github.com/jimweirich/builder/blob/master/test/performance.rb#L17). If a project uses `track_files "**/*.rb"` and has the builder gem in a subdir such as vendor, SimpleCov will fail even though the project code is perfectly valid.

This fixes a regression that was introduced in v0.15.0 when `add_not_loaded_files` was changed to call `LinesClassifier#classify`, which runs a RegExp on each line--and raises ArgumentError if the line had an invalid byte sequence.

### DIFF
--- a/lib/simplecov/lines_classifier.rb
+++ b/lib/simplecov/lines_classifier.rb
@@ -20,12 +20,17 @@ module SimpleCov
       skipping = false
 
       lines.map do |line|
-        if line =~ self.class.no_cov_line
-          skipping = !skipping
-          NOT_RELEVANT
-        elsif skipping || line =~ WHITESPACE_OR_COMMENT_LINE
-          NOT_RELEVANT
-        else
+        begin
+          if line =~ self.class.no_cov_line
+            skipping = !skipping
+            NOT_RELEVANT
+          elsif skipping || line =~ WHITESPACE_OR_COMMENT_LINE
+            NOT_RELEVANT
+          else
+            RELEVANT
+          end
+        rescue ArgumentError
+          # E.g., line contains an invalid byte sequence in UTF-8
           RELEVANT
         end
       end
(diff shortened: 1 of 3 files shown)


lib/simplecov/lines_classifier.rb
```diff
@@ -20,12 +20,17 @@ def classify(lines)
...
+          if line =~ self.class.no_cov_line
+            skipping = !skipping
+            NOT_RELEVANT
+          elsif skipping || line =~ WHITESPACE_OR_COMMENT_LINE
+            NOT_RELEVANT
+          else
+            RELEVANT
+          end
+        rescue ArgumentError
+          # E.g., line contains an invalid byte sequence in UTF-8
```
PragTob: Good explanatory why comment!

PragTob:
Ok, so I won't fix this now... maybe later. Problem is the `skipping` variable that relies on the local scope. I guess we could make it an instance variable. Currently evaluating options.

Problem is also that the same logic is [applied elsewhere](https://github.com/colszowka/simplecov/blob/master/lib/simplecov/source_file.rb#L181-L192) - best we'd put that logic inside the lines classifier so there's no need to duplicate. Question is how to best do that. Currently not coming up with a nice lightweight abstraction to suit both cases. I guess we could use blocks. But wanna check the performance of that, one of the reasons I got #665 - that helps us evaluate if it creates a performance problem or not.

BMorearty:
Good point that the same logic is applied elsewhere.

I'm motivated to get this fixed (it's a must-have for us so I need to either fix it here or fork the project for Airbnb, which I'd rather not do). I'm trying out different solutions so you don't have to spend too much time on this. I know open-source maintenance is time-consuming.

... later ...

I thought about using `if (line =~ self.class.no_cov_line rescue true)` but it's less clear and still doesn't make Rubocop happy. 

I also attempted to refactor the two places that use this shared logic into one but I don't see any way to do it. There are too many logic differences:

1. One checks for `WHITESPACE_OR_COMMENT_LINE` and one does not.
2. One has a final `else` clause and one does not.

I've got an alternate solution in another branch that Rubocop is ok with but does not attempt to share logic between the two methods since they are too different. I'll make a separate PR.

PragTob:
super seeded by #669

PR #669, opened 2018-03-10
Ignore invalid byte sequences

BMorearty:
This is an improvement on #664, which fixes SimpleCov so it won't crash when it encounters an invalid byte sequence in UTF-8.

This version is better in these ways:

1. The `classify` method is still short enough to make Rubocop happy.
2. The code that ignores invalid byte sequences is reused in the two places where we iterate over all the lines looking for `:nocov:`: `SimpleCov::LinesClassifier#classify` and `SimpleCov::SourceFile.process_skipped_lines`.

Like #664, this fixes a regression that was introduced in v0.15.0 when `add_not_loaded_files` was changed to call `LinesClassifier#classify`, which runs a RegExp on each line--and raises ArgumentError if the line had an invalid byte sequence.


BMorearty:
Not sure why Cucumber is failing in Ruby 1.9.3. "It works on my machine."

PragTob:
That's the flaky I tried to fix and obviously failed... sorry :|

PR #670, opened 2018-03-11
No need to ship our specs and cukes :)

PragTob:


simplecov.gemspec
```diff
@@ -22,8 +22,6 @@ Gem::Specification.new do |gem|
 
   gem.add_development_dependency "bundler", "~> 1.9"
 
-  gem.files         = `git ls-files`.split("\n") - %w[cucumber.yml]
-  gem.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n")
-  gem.executables   = `git ls-files -- bin/*`.split("\n").map { |f| File.basename(f) }
```
PragTob: we have no executables, test_files seems to be deprecated....

PR #672, opened 2018-03-15
Make sure the license file gets included with the gem

tas50:
When the list of files to include was pruned down the LICENSE file was
accidently pruned as well. This causes issues for projects that do
strict license validation on depdendencies.

Signed-off-by: Tim Smith <tsmith@chef.io>


tas50:
I didn't update the changelog because I wasn't sure what you'd want it to look like, but we'd love if this went out as 0.16.1.

PragTob:
Sorry that happened!

I think renaming the file to `LICENSE` is more in line with community standards so that'd be my preferred change :)

tas50:
@PragTob Updated to use LICENSE

PragTob:
@tas50 and released as 0.16.1

PR #675, opened 2018-04-20
Update README.md

kirylpl:
Adding terminal command for updating gitignore

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,11 @@ Getting started
    are not tracked by Git (optional):
 
     ```
-    coverage
+    echo "coverage" >> .gitignore
+    ```
+    Or if you use Windows:
+    ```
+    echo coverage >> .gitignore
     ```
 
     If you're making a Rails application, SimpleCov comes with built-in configurations (see below for information on profiles)


bf4:
@kirylpl Do you know if that works for everyone, including windows users?

kirylpl:
@bf4 yes, echo is cross-platform command https://en.wikipedia.org/wiki/Echo_(command)

bf4:
And >> ?

B mobile phone

kirylpl:
@bf4 You were right. There is no need in comma on windows. Updated.

README.md
```diff
@@ -87,7 +87,11 @@ Getting started
    are not tracked by Git (optional):
 
     ```
-    coverage
+    echo "coverage" >> .gitignore
+    ```
+    Or if you use Windows for some reason:
```
PragTob: I think we should get rid of the `for some reason` part - no need to belittle Windows users.

kirylpl: @PragTob Proudly speaking, never met ruby developer using Windows. So, imo, it will not touch anybody.

PragTob: I know quite some of them also including popular Rubyists :)

kirylpl: @PragTob Got it, I add some row for this fanatics also :)

README.md
```diff
@@ -87,7 +87,11 @@ Getting started
    are not tracked by Git (optional):
 
     ```
-    coverage
+    echo "coverage" >> .gitignore
+    ```
+    Or if you use Windows:
+    ```
+    echo coverage >> .gitignore
```
bf4: am I missing something or is the only difference the quotes? If so, you can omit quotes on *nix as well. 

My own feeling is that if you can't be trusted to know how to add a line to your gitignore you have bigger issues.

PR #676, opened 2018-05-02
please capybara dont fail on me

PragTob:
for reasons unbeknownst to me cucumber won't run on my local setup and just wait forever... sigh.

Let's see if this fixes the red builds.


scmx:
The test suite fails for me in master, but I got it running locally by adding puma to Gemfile instead
```diff
  group :development do
    gem "benchmark-ips"
+   gem "puma"
  end
```
I was going to open a PR for that, but then I found yours instead which seems to do the trick just as good.

PR #684, opened 2018-07-25
Add a default_group_name option to override "Ungrouped"

Kjarrigan:
Using groups is a great way to make the result more readable but I always have some "Remaining" files which I did not want to group but I dislike the "Ungrouped" keyword. So I looked through the code and it was a rather easy, isolated "fix".

I know I could achieve the same thing with the current setup with a Regex-filter that negates all other filtes but this even uglier then the "Ungrouped" word.

Before:
```ruby
SimpleCov.start do
  add_group 'Foo', 'foo/'
end
# | All Files | 100.0% | 2 Files |
# | Foo       | 100.0% | 1 File  |
# | Ungrouped | 100.0% | 1 File  |

```

After:
```ruby
SimpleCov.start do
  default_group_name 'Remaining'
  add_group 'Foo', 'foo/'
end
# | All Files | 100.0% | 2 Files |
# | Foo       | 100.0% | 1 File  |
# | Remaining | 100.0% | 1 File  |
```

One thing I noticed was that `@test_unit` and `@rspec` of the `groups_and_filters_complex.feature` were different (test_unit was missing the `By array` filter). Was this on purpose? Now both tests are exactly the same.

### DIFF
--- a/features/rspec_groups_and_filters_complex.feature
+++ b/features/rspec_groups_and_filters_complex.feature
@@ -11,6 +11,8 @@ Feature: Sophisticated grouping and filtering on RSpec
       """
       require 'simplecov'
       SimpleCov.start do
+        default_group_name 'Remaining'
+
         add_group 'By block' do |src_file|
           src_file.filename =~ /MaGiC/i
         end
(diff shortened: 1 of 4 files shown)


Kjarrigan:
@PragTob  any thoughts?

Kjarrigan:
@colszowka Is this project still alive or do I have to seek for another coverage tool?

tonobo:
Any news?

PR #687, opened 2018-08-07
Support multiple files in Result.from_hash

sheki:
Result.from_hash is used to merge resultsets when the tests are run in
parallel across machines using things like `parallel_tests`. In the
above scenario each machine can have multiple results sets as one would
want to run as many parallel spec process as the machine has cores. The
merge result code only consumes the first result, and dropping the rest.

In this PR I fix it by returning arrays through out.

### DIFF
--- a/lib/simplecov/result.rb
+++ b/lib/simplecov/result.rb
@@ -66,11 +66,15 @@ module SimpleCov
 
     # Loads a SimpleCov::Result#to_hash dump
     def self.from_hash(hash)
-      command_name, data = hash.first
-      result = SimpleCov::Result.new(data["coverage"])
-      result.command_name = command_name
-      result.created_at = Time.at(data["timestamp"])
-      result
+      results = []
+      hash.each do |command_name, data|
+        command_name, data = hash.first
+        result = SimpleCov::Result.new(data["coverage"])
+        result.command_name = command_name
+        result.created_at = Time.at(data["timestamp"])
+        results << result
+      end
+      results
     end
 
   private
--- a/lib/simplecov/result_merger.rb
+++ b/lib/simplecov/result_merger.rb
@@ -51,15 +51,17 @@ module SimpleCov
       # All results that are above the SimpleCov.merge_timeout will be
       # dropped. Returns an array of SimpleCov::Result items.
       def results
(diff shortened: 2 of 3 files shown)


lib/simplecov/result.rb
```diff
@@ -66,11 +66,15 @@ def to_hash
...
     # Loads a SimpleCov::Result#to_hash dump
     def self.from_hash(hash)
-      command_name, data = hash.first
-      result = SimpleCov::Result.new(data["coverage"])
-      result.command_name = command_name
-      result.created_at = Time.at(data["timestamp"])
-      result
+      results = []
+      hash.each do |command_name, data|
+        command_name, data = hash.first
```
PragTob: this seems odd to me, we iterate over it getting `command_name` and `data` and then we just discard them and use `hash.first` again - shouldn't this be prone to the same problem(s) again?

lib/simplecov/result.rb
```diff
@@ -66,11 +66,15 @@ def to_hash
...
+        result = SimpleCov::Result.new(data["coverage"])
+        result.command_name = command_name
+        result.created_at = Time.at(data["timestamp"])
+        results << result
```
PragTob: if we just add results shouldn't this rather be a `map` operation then?

PragTob:
So, I finally resumed this and the reasoning why this works (it'd have been hard core borken otherwise) is because the iteration happens at another level. Namely over here:

https://github.com/simplecov-ruby/simplecov/blob/67ced4a0639ef09800d7de528a74be828a53cda5/lib/simplecov/result_merger.rb#L55-L63

Now you could make an argument that all of this should happen in `.from_hash` and be called from those places instead. 

It's also meant as an inverse operation to `to_hash`:

https://github.com/simplecov-ruby/simplecov/blob/67ced4a0639ef09800d7de528a74be828a53cda5/lib/simplecov/result.rb#L63-L74

I'll meditate on this a bit and see how I like it. Maybe sketch out the solution for moving it up or alternatively probably put some sort of "raise assertion" in there.

Closing here as it'll probably not look like in here/not implemented at all. Thanks a lot for the pointer though :green_heart:

PR #688, opened 2018-08-10
Explicitly print the exit status when SimpleCov fails the build

daemonsy:
## What's up? 

My team uses SimpleCov at work. Thank you for the work. We have a Jenkins setup that runs specs using [parallel_tests](https://github.com/grosser/parallel_tests). 

Recently we  turned on `SimpleCov.minimum_coverage` and sometimes we started getting random failures. Eventually, we found out that it was because of the way minimum coverage is calculated, certain `parallel_tests` groupings had coverage below the minimum. We got final output like this: 

> 13752 examples, 0 failures, 120 pendings
> Tests have failed for a parallel_test group.

With parallel_tests and our Jenkins setup, the message about "expected minimum coverage..." wasn't at the end, so it's isn't clear. 

During the debugging process, I spent some time looking for words like `failed`, `failure`, `exit` to try to see if the build was failed by a non exit 0. 

## What this does

With this, I thought it might be a good addition to SimpleCov to explicitly print out exit statuses and that it is failing the build. When there is a non-zero exit, this message is printed during the `run_exit_tasks!` method. 

> SimpleCov failed build with exit <status> 

It would have helped me discover earlier that SimpleCov was exiting non zero.  So hopping that this default message will help other users spot the cause of an exit earlier.

### DIFF
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -203,7 +203,10 @@ module SimpleCov
 
       # Force exit with stored status (see github issue #5)
       # unless it's nil or 0 (see github issue #281)
-      Kernel.exit exit_status if exit_status && exit_status > 0
+      if exit_status && exit_status > 0
+        $stderr.printf("SimpleCov failed build with exit %d", exit_status)
+        Kernel.exit exit_status
+      end
     end
 
     # @api private
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -204,7 +204,7 @@ module SimpleCov
       # Force exit with stored status (see github issue #5)
       # unless it's nil or 0 (see github issue #281)
       if exit_status && exit_status > 0
-        $stderr.printf("SimpleCov failed build with exit %d", exit_status)
+        $stderr.printf("SimpleCov failed with exit %d", exit_status)
         Kernel.exit exit_status
       end
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+<Version> <Date>
+===================
+
+## Enhancements
+
+* Print the exit status explicitly when it's not a successful build so it's easier figure out SimpleCov failed the build in the output.
+
+
 0.16.1 (2018-03-16)
 ===================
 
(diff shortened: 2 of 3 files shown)


lib/simplecov.rb
```diff
@@ -203,7 +203,10 @@ def run_exit_tasks!
 
       # Force exit with stored status (see github issue #5)
       # unless it's nil or 0 (see github issue #281)
-      Kernel.exit exit_status if exit_status && exit_status > 0
+      if exit_status && exit_status > 0
+        $stderr.printf("SimpleCov failed build with exit %d", exit_status)
```
daemonsy: Since this affects all failure modes, it should probably be tested at a more fundamental level than the `minimum_coverage.feature` cuke. Will be glad to take advice from maintainers on testing this

PragTob: Not sure I like the wording here - it's not always a build I run simplecov locally all the time. How about we remove the `build` and jsut say `SimpleCov failed with exit %d` ?

PragTob: I don't think we need to test all permutations, that we hit this branch once and that it prints seems a good enough test for me :+1:

PragTob:
@daemonsy I think you forgot to adjust the test with the new message :)

daemonsy:
😅 we know the tests work. Updated them. 

If you feel this is good enough to merge, feel free to squash the commits

daemonsy:
Just a ping @PragTob in case the previous message didn't trigger a notification

daemonsy:
Hello, just a bump on this.

daemonsy:
Hi @PragTob just doing my rounds to clear pending OS PRs.

daemonsy:
Hi, anyone got a chance to look at this yet? thanks :pray:

daemonsy:
Hi hi, just one last bump on this.

PragTob:
:wave: 

Hi @daemonsy - sorry I took a bit of a break from simplecov as there was a lot going on in my life et. al.

Also you're not an idiot (commit message) you're a nice person helping in open source! :green_heart:

daemonsy:
No problem @PragTob ! I hope everything is going well. 
simplecov is still helping us on every, single, build, so thank you. 

(P.S. thanks for the nice words, one can be both at the same time :P)

lib/simplecov.rb
eamonn-webster: A line feed would have been prettier

Issue #686, opened 2018-08-01
TextFormatter like istanbul's text coverage report


tansaku:
I've searched through the issues and found a couple of mentions of istanbul, but no one sharing the particular text output format it uses, e.g.

```
----------------|----------|----------|----------|----------|----------------|
File            |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------------|----------|----------|----------|----------|----------------|
All files       |    79.41 |     57.5 |    83.78 |    79.41 |                |
 async-voter.js |     88.1 |    77.78 |      100 |     88.1 |... 49,75,76,77 |
 index.js       |       95 |    83.33 |       50 |       95 |             19 |
 persist.js     |    46.88 |       25 |       50 |    46.88 |... 55,69,77,79 |
----------------|----------|----------|----------|----------|----------------|
```

I find this quick and simple text output at the end of a test run to be super useful.  Is there an existing textformatter that I can use with simplecov that will output this kind of data on the command line after a test run?

A bit like the [rspec --profile option](https://stackoverflow.com/a/6946446/316729) does?

Maybe this will require writing my own formatter?  presumably simplecov can output data to the command line - not just static files?

Many thanks in advance

PragTob:
:wave: 

An overview of existing formatters can be found here: https://github.com/colszowka/simplecov/blob/master/doc/alternate-formatters.md

Failing at that, yes you'd need your own formatter or we could make an official simplecov text formatter - quite honestly I don't have much time allocated for simplecov these days though.

tansaku:
@PragTob totally understand about not having much time :-) happy to write a formatter if you can point me at anything that will help me get started - are there any docs on making a formatter, or is it just a question of looking through the existing formatters to get a feel for how to create one?

PragTob:
@tansaku best to look at how the html formatter handles it, as it's by far the most used/tested one :)

tansaku:
ah, looks like someone did it already https://github.com/chetan/simplecov-console

PR #691, opened 2018-08-14
add an IstanbulFormatter so it gives output on a simple project

tansaku:
This is just preliminary take on how an IstanbulFormatter might be a variation on the SimpleFormatter so that it displays something on a simple project without having to specify any groups up front.  I was finding simplecov would generate no command line output with SimpleFormatter for small or large projects unless I explicitly added groups.

These changes allow a project to select the InstanbulFormatter and get output like this on the command line, which is a step towards an instanbul style output:

```
→ rspec
required simplecov
.

Finished in 0.00354 seconds (files took 0.11997 seconds to load)
1 example, 0 failures

Coverage Output
========================================
code.rb (coverage: 66.67%)
test_spec.rb (coverage: 100.0%)
```

I'd like to explore making this look even more like the instanbul output, but this is a start.  Other things to include are:

* [ ] colorizing output red/green based on some threshold coverage
* [ ] uncovered line numbers
* [ ] separating out %stmts, %branches, %methods, %lines (already supported in simplecov?)
* [ ] other colors to indicate partial coverage (light yellow in instanbul)

I haven't added any tests yet as I'm just getting started and trying to understand if this approach makes sense, but I think I've just worked out how to create some tests.  I'll add those and also think about whether this shouldn't be a standalone formatter in a separate repo.

related to #686

### DIFF
--- a/lib/simplecov/formatter.rb
+++ b/lib/simplecov/formatter.rb
@@ -7,4 +7,5 @@ module SimpleCov
 end
 
 require "simplecov/formatter/simple_formatter"
+require "simplecov/formatter/istanbul_formatter"
 require "simplecov/formatter/multi_formatter"
--- a/lib/simplecov/formatter/instanbul_formatter.rb
+++ b/lib/simplecov/formatter/instanbul_formatter.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+#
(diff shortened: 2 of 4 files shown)


tansaku:
ah, looks like someone has already done what I wanted: https://github.com/chetan/simplecov-console

PR #705, opened 2018-12-08
Fix CI for ruby-head to be run

yui-knk:
We can not specify "trunk" for `#platforms` of Gemfile, so we can not run CI with "ruby-head".
This PR separates each gemfiles to 'gemfiles/*.gemfile'. We can specify arbitrary gem versions by this change.


amatsuda:
@yui-knk Nice work!

yui-knk:
Thanks!

PR #707, opened 2018-12-26
Delete condition for ActionCable group in Rails profile.

koheisg:
ActionCable is enabled in a default Rails project. (There may be few projects using ActionCable...)
In general, simplecov reads before Rails starts up, so ActionCable is not defined at this time.
That is, since the group in the ActionCable is not always true, I submitted this pull request.


PragTob:
:wave: 

Hey yeah good thing to notice that it can't be defined :blush: I see no other way to easily check this - that it's potentially empty is probably better fixed somewhere else.

Thanks!

PR #708, opened 2019-01-17
No need to explicitly bundle gems included in the gemspec

deivid-rodriguez:
This PR restores https://github.com/colszowka/simplecov/commit/0d2fe5b543060335fb6f234421d7a21e45ea98f1, with a small modification: specify `development_group: 'test'` as an argument to `gemspec`, so that it works properly on TravisCI.

### DIFF
--- a/gemfiles/jruby.gemfile
+++ b/gemfiles/jruby.gemfile
@@ -23,4 +23,4 @@ group :test do
   gem "json", "~> 2.0"
 end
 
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
Thoughts, @amatsuda?

amatsuda:
@deivid-rodriguez Wow, nice!!! Thank you so much! 😻 😻 😻

PR #711, opened 2019-01-27
Fix missing delegator for Result#lines_of_code

kputnam:
I think because the line read `def_delegator :files, :lines_of_code, :total_lines` instead of def_delegator**s**, this caused a NoMethodError when calling `Result#lines_of_code`.


kputnam:
I'm not sure why this worked on my machine but not on the build server, maybe I didn't check in all the changes. Never mind, I'll close this!

PR #714, opened 2019-02-24
README: Drop defunct Gemnasium badge

olleolleolle:
This PR removes the badge for the defunct service Gemnasium.


colszowka:
Ah, good catch! Thanks for the patch!

PR #717, opened 2019-03-25
Support project roots with special characters

deivid-rodriguez:
I was configuring simplecov in bundler's specs and I got this crash in one of the specs:

```
       /home/travis/build/bundler/bundler/tmp/gems/base/gems/simplecov-0.16.1/lib/simplecov/source_file.rb:87:in `project_filename': empty char-class: /^\/home\/travis\/build\/bundler\/bundler\/tmp\/foo[]bar/ (RegexpError)
```

This PR should fix that super-edge case.


amatsuda:
@deivid-rodriguez 👍

PR #720, opened 2019-04-04
CI: Drop unused Travis option

olleolleolle:
Removes an old setting from Travis. See https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

I marked this as CI skip - perhaps that's not how to do it, in this project, but the setting "does nothing", now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ before_install:
 
 bundler_args: --without development --jobs=3 --retry=3
 
-sudo: false
-
 matrix:
   include:
     - rvm: 1.8.7


olleolleolle:
@PragTob This, a very small ✨ change (cleanup).

PragTob:
Whew didn't know, thank you!

Normally I'd rather have it without CI-skip just to make sure we didn't break the CI but you're right if travis says it has no effect I'll trust them. 

This also means I'll have to remove the sudo required from all my Elixir projects but that's OT :D

PragTob:
@olleolleolle Not sure what that means but I'm happy to have it :grin:

PR #721, opened 2019-04-08
Add filter for hidden files and folders per default

schmijos:
It seems a bit weird, that hidden files and folders are included in the coverage measurement per default. They are hidden for a reason.

I now filter per default hidden folders and files in the root folder.

@colszowka do you have a strong opinion about that?

### DIFF
--- a/lib/simplecov/profiles/hidden_filter.rb
+++ b/lib/simplecov/profiles/hidden_filter.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+SimpleCov.profiles.define "hidden_filter" do
+  add_filter %r{^/\..*}
+end
--- a/spec/filters_spec.rb
+++ b/spec/filters_spec.rb
@@ -173,6 +173,18 @@ if SimpleCov.usable?
         it "filters vendor/bundle" do
           expect(SimpleCov.filtered([a_file("vendor/bundle/foo.rb")]).count).to eq(0)
         end
+
+        it "filters hidden folders" do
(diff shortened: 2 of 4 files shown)


lib/simplecov/profiles/hidden_filter.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+SimpleCov.profiles.define "hidden_filter" do
+  add_filter %r{^/\..*}
```
schmijos: I could as well imagine to filter all hidden files and folders and not only the ones in the root folder. I didn't do it because it could be harder to debug.

spec/filters_spec.rb
```diff
@@ -173,6 +173,14 @@ def a_file(path)
...
           expect(SimpleCov.filtered([a_file("vendor/bundle/foo.rb")]).count).to eq(0)
         end
+
+        it "filters hidden folders" do
+          expect(SimpleCov.filtered([a_file(".semaphore-cache/lib.rb")]).count).to eq(0)
+        end
+
+        it "filters hidden files" do
+          expect(SimpleCov.filtered([a_file(".hidden_config.rb")]).count).to eq(0)
+        end
```
PragTob: Could you please add a test to verify that we're not filtering hidden sub folders? I like your argument for not filtering so I think we should spec it as expected behaviour :)

schmijos: You're totally right! I added the test.

PR #729, opened 2019-07-02
Remove ruby 2.3 support but add JRuby 9.1 back in

PragTob:


PragTob:
JRuby on travis via rvm seems to fail almost all the time... had to restart like 3 times :(

PR #731, opened 2019-07-10
Add method to shutdown and collect results before at_exit 

macumber:
We are using simplecov as part of a testing framework and have need to collect results before `at_exit`.  The new `end_now` method allows the user to get results before `at_exit`.


macumber:
@colszowka I can add a test for this new method, wanted to check if you were ok with the general premise first though

PragTob:
:wave: 

Hi there,

sorry for the delay I've been rather busy :) What's you use case if I might ask? It's rather specific. To understand if there'd be a better way to do this/accomplish this I need to understand the use case.

lib/simplecov.rb
```diff
@@ -53,6 +53,17 @@ def start(profile = nil, &block)
...
     end
 
+    #
+    # Method added by NREL to shutdown SimpleCov and collect results
+    # this will set running to false which will prevent from running again at_exit
+    def end_now
+      SimpleCov.set_exit_exception
+      exit_status = SimpleCov.exit_status_from_exception
+      SimpleCov.result.format!
+      exit_status = SimpleCov.process_result(SimpleCov.result, exit_status)
```
PragTob: I think `SimpleCov` should be `self` here so no need to address it directly imo. Also I think the last assignment and return aren't needed :)

lib/simplecov.rb
```diff
@@ -53,6 +53,17 @@ def start(profile = nil, &block)
       Coverage.start
     end
 
+    #
+    # Method added by NREL to shutdown SimpleCov and collect results
```
PragTob: What is NREL?

PragTob:
@macumber Hi there! :wave: Any plans on picking this up or detailing your use case for this? Would be really helpful! :)

deivid-rodriguez:
Closing this because it's being included in #780. @macumber the linked PR achieves the same thing you were looking for here. Only difference is that the "early exit method" is called `SimpleCov.run_exit_tasks!`, not `SimpleCov.end_now`.

PR #733, opened 2019-07-16
[nit] clear on how to open page

chiwenchen:


PragTob:
Thanks but we already have something similar now! :tada:

PR #735, opened 2019-07-23
Fix the Coverage Link

tchia04:
Make sure the Coverage link is pointing to the latest version instead of a hard coded version which will get stale overtime.


colszowka:
+https

Great, thank you!

PR #737, opened 2019-07-26
Warn the user if the minimum coverage is set greater than 100

belfazt:
I was trying something out, and I thought that it would be funny/useful to point this out to the user

### DIFF
--- a/lib/simplecov/configuration.rb
+++ b/lib/simplecov/configuration.rb
@@ -224,6 +224,7 @@ module SimpleCov
     # Default is 0% (disabled)
     #
     def minimum_coverage(coverage = nil)
+      minimum_possible_coverage_exceeded("minimum_coverage") if coverage && coverage > 100
       @minimum_coverage ||= (coverage || 0).to_f.round(2)
     end
 
@@ -245,6 +246,7 @@ module SimpleCov
     # Default is 0% (disabled)
     #
     def minimum_coverage_by_file(coverage = nil)
+      minimum_possible_coverage_exceeded("minimum_coverage_by_file") if coverage && coverage > 100
       @minimum_coverage_by_file ||= (coverage || 0).to_f.round(2)
     end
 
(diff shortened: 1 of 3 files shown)


lib/simplecov/configuration.rb
```diff
@@ -288,6 +290,10 @@ def add_group(group_name, filter_argument = nil, &filter_proc)
 
   private
 
+    def minimum_possible_coverage_exceeded(method_name)
+      warn "The coverage you set for #{method_name} is greater than 100%"
```
PragTob: I think we should rather call this `coverage_option` or something (still doesn't sound great) - _method_name_ makes me think we're doing fancy metaprogramming stuff :)

spec/configuration_spec.rb
```diff
@@ -33,5 +33,19 @@
         expect(config.tracked_files).to be_nil
       end
     end
+
+    describe "#minimum_coverage" do
+      it "warns you about your usage" do
+        expect(config).to receive(:warn).with("The coverage you set for minimum_coverage is greater than 100%")
+        config.minimum_coverage(100.01)
```
PragTob: Thanks for the tests! We're missing a negative test though. Aka 100.0 doesn't trigger a warning (current tests would pass if we just warned all the time it seems like to me)

PragTob:
Ugh just saw the JRuby failure, I'd guess/hope that it's not related I'll rerun that and otherwise will have to do some testing.

belfazt:
Yeah, I don't know what could've gone wrong with JRuby

PragTob:
failure in JRuby is persistent... I'll open a mock PR without changes to see if that also fails and then try to fix it. Sorry for the hassle :(

belfazt:
It seems that rvm is not identifying `9.1.8.0`?

PragTob:
9.2.8.0 yeah, my bad as charlie commented it's fixed there I just tried it but it's not released. I set the JDK version instead. Works, I'll be brave and emrge this :)

belfazt:
Ah, sorry, hehe. Thank you very much!

PR #738, opened 2019-07-27
Upgrade JRuby

PragTob:
https://github.com/jruby/jruby/issues/5736


PragTob:
Ofc it wasn't released what did I think... will try to pin the jdk

PR #739, opened 2019-08-05
Change misleading wording in exit message

zarqman:
PR #688 added a message when the exit status != 0. However, when the exit status is non-zero because a test failed, the message wording is misleading because it specifically refers to SimpleCov. I personally interpreted it as an error in SimpleCov's report generation, when SimpleCov was actually working just fine!

This PR makes the wording more generic so as to not mislead one into thinking SimpleCov is misbehaving when it isn't. 😄 

Also, the message was missing the trailing "\n", so added that as well.

Old output:
```
SimpleCov failed with exit 1$
```

New output:
```
Exiting with status 1
$
```

### DIFF
--- a/features/minimum_coverage.feature
+++ b/features/minimum_coverage.feature
@@ -17,7 +17,7 @@ Feature:
     When I run `bundle exec rake test`
     Then the exit status should not be 0
     And the output should contain "Coverage (88.10%) is below the expected minimum coverage (90.00%)."
-    And the output should contain "SimpleCov failed with exit 2"
+    And the output should contain "Exiting with status 2."
 
   Scenario:
     Given SimpleCov for Test/Unit is configured with:
@@ -32,7 +32,7 @@ Feature:
     When I run `bundle exec rake test`
     Then the exit status should not be 0
     And the output should contain "Coverage (88.10%) is below the expected minimum coverage (88.11%)."
-    And the output should contain "SimpleCov failed with exit 2"
+    And the output should contain "Exiting with status 2."
 
   Scenario:
     Given SimpleCov for Test/Unit is configured with:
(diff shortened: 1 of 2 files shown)


PragTob:
:wave: 

Hi there, thanks for contributing!

I'm not sure we should drop simplecov from the message but rather improve the message as I'd wanna know where the message comes from.

Something like SimpleCov made the process exit with `$reason` or something :)

zarqman:
Sure, and yet that's just the thing--SimpleCov is also intercepting exit statuses from earlier up (eg: a test failure), and is then claiming to have generated/caused the error.

When SimpleCov actually causes the error, then I agree, it's ideal for SimpleCov to identify itself as the source of the non-zero exit. However, when SimpleCov is intercepting a non-zero status originating elsewhere, it should **not** then claim that as its own.

Perhaps it'd be better to entirely remove the message added in #688, and then add separate messages elsewhere in SimpleCov when SimpleCov itself sets the exit status to non-zero based on its own conditions (ie: insufficient test coverage)?

PragTob:
@zarqman I think it's cool for simplecov to identify itself as the originator of the message but you are right it could be more explicit like "a process failed with status x" and otherwise being more explicit about that simplecov set the exit status due to reason y I agree about that. 

Thinking about it... I agree, maybe only if we set the exit status ourselves due to our logic and not because it bubbled up, good idea @zarqman !

PragTob:
#742 will cause a small merge conflict for you but I think it's fine :)

PragTob:
@zarqman any plans on pursuing this or shall I take over?

zarqman:
haven't had a chance to get back to this...sorry! feel free to take it over.

PragTob:
Taken the new approach to #906

PR #742, opened 2019-08-08
Add new line to exit_status log

harpocrates12:
Add new line to exit_status log.


PragTob:
Ugh, good point thank you very much for the contribution! :tada:

PR #746, opened 2019-09-11
Avoid hanging with parallel_tests

annaswims:
`ParallelTests.last_process?` returns `true` if the process is the last one to start.  We are
interested in the last test to finish so we want to use `ParallelTests.number_of_running_processes`

Because it's possible to have the parallel_tests gem in your project and still run tests serially we
want to also check if tests are currently being run in parallel.  If `ENV['TEST_ENV_NUMBER']` is defined we know tests are currently being run in parallel.

### DIFF
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -236,7 +236,8 @@ module SimpleCov
     # @api private
     #
     def final_result_process?
-      !defined?(ParallelTests) || ParallelTests.last_process?
+      # checking for ENV["TEST_ENV_NUMBER"] to determine if the tess are being run in parallel
+      !defined?(ParallelTests) || !ENV["TEST_ENV_NUMBER"] || ParallelTests.number_of_running_processes <= 1
     end
 
     #


mcmoyer:
applying this fix is allowing me to run parallel specs with SimpleCov v0.17.0 again

annaswims:
Glad to hear it @mcmoyer! By any chance is your coverage report correct?  My tests don't hang any more, but the coverage report is way low and I haven't had a chance to look into it yet.

mcmoyer:
@annaswims - my coverage report appears correct.

deivid-rodriguez:
This seems to have fixed some hangs in the activeadmin's test suite as well.

PragTob:
New version is released: https://rubygems.org/gems/simplecov/versions/0.17.1

PragTob:
There was an open issue, I missed it. Sorry everyone :disappointed:

mcmoyer:
heck, I’m just happy to see it addressed so quickly.  Thanks for the quick turn around.

PragTob:
@mcmoyer sure, thanks :) However it's sad that it was in the wild for 2.5 months, one of the drawbacks of not working on a big ruby app where I'd test out every release immediately any more. Plus reduced OSS time :| I'm glad I checked in today.

deivid-rodriguez:
Thanks everyone for fixing and releasing this! Will upgrade activeadmin test suite to use 0.17.1, and get back here after I confirm the issue is really gone.

deivid-rodriguez:
We did get a jruby crash after the upgrade, but haven't seen any hangs while using this code. My guess is that the crash was previously happening too, but being hidden by the hang. So I think this PR should be good, but I'll update again if I see any issues.

PragTob:
@deivid-rodriguez thanks! You might want to report the crash to our JRuby friends so that they can have a look :)

deivid-rodriguez:
Yes, I'll definitely report it! :)

PR #756, opened 2019-11-11
Fix at_exit hook to cooperate with Minitest

adam12:
Ensures that the `at_exit` hook is triggered at the right time when run under Minitest, by using the Minitest `after_run` hook for execution.

Basically verbatim of @zenspider's [suggested changes.](https://github.com/colszowka/simplecov/issues/685#issuecomment-499279291)


bf4:
Fine with me. I never got minitest to work satisfactorily with simplecov. Sounds like you did, so kudos

bf4:
Would be nice to have some sort of regression test on this, though I'm not quite sure how.  

start a new process? and shell out to a file the opens with a required minitest/autorun ?

adam12:
> Would be nice to have some sort of regression test on this, though I'm not quite sure how.

Leave it with me. Maybe we can shell out and test the Simplecov output matches an expected coverage.

adam12:
> Fine with me. I never got minitest to work satisfactorily with simplecov

Ironically once I've started testing it I can't seem to get it to reproduce. Running it against a real test suite it's fine tho.

Going to keep looking at it.

spec/defaults_spec.rb
```diff
@@ -38,4 +38,11 @@ def filtered?(config, filename)
       expect(filtered?(config, "config/environment.rb")).to be
     end
   end
+
+  context "exit hook" do
+    it "uses Minitest if available" do
+      expect { system RbConfig.ruby, "spec/fixtures/frameworks/minitest.rb" }
+        .to output(Regexp.new(Regexp.escape("4 / 5 LOC (80.0%) covered"))).to_stdout_from_any_process
```
PragTob: Hey, thanks for adding a test! :+1: 

However, this looks like it might have a better home in the cucumber integration tests that basically run tests like these, have matchers etc. Care to move it over there maybe? :)

adam12: 100%. I didn't care for it's current location but wasnt sure where else to put it. Let me work on moving it to Cucumber.

PragTob: if you need help feel free to ping me

adam12: I've run into a few interesting issues that's resulted in some yak shaving:

1. Phantomjs is EOL, and as such, will no longer run on FreeBSD 12. I've swapped it temporarily with Cuprite (a Chrome driver for Capybara) and it's satisfied Capybara but this leads me to the next issue.
2. Without changing anything other than the Capybara driver, I have multiple features failing due to the output of the commands being run. They are naturally passing on Travis so there's some confusion. Specs are green.

If you have a `Gemfile.lock` where the features are green that you could provide, at least I could rule out dependency mismatch.

PragTob: @adam12 ugh... I'm so sorry for the troubles. Worst case I'll "translate" the tests to cucumber for you don't want to have you go through a huge amount of it... let me run all of it locally and if passing upload my Gemfile.lock.

PragTob: You can see working dep versions in the travis run... I just had a problem myself as I had an outdated cucumber version. I'll adjust dependency versions and PR. 

All this makes me thing of José Valim never understanding/liking that libraries in ruby didn't use the `.lock` file to avoid problems like this.

PragTob: Ok here's a working `Gemfile.lock` for me to run the tests at master:

<details>

```
PATH
  remote: .
  specs:
    simplecov (0.18.0.pre.dev)
      docile (~> 1.1)
      simplecov-html (~> 0.10.0)

GEM
  remote: https://rubygems.org/
  specs:
    addressable (2.7.0)
      public_suffix (>= 2.0.2, < 5.0)
    aruba (0.14.11)
      childprocess (>= 0.6.3, < 3.0.0)
      contracts (~> 0.9)
      cucumber (>= 1.3.19)
      ffi (~> 1.9)
      rspec-expectations (>= 2.99)
      thor (~> 0.19)
    ast (2.4.0)
    backports (3.15.0)
    benchmark-ips (2.7.2)
    builder (3.2.3)
    capybara (2.18.0)
      addressable
      mini_mime (>= 0.1.3)
      nokogiri (>= 1.3.3)
      rack (>= 1.0.0)
      rack-test (>= 0.5.4)
      xpath (>= 2.0, < 4.0)
    childprocess (2.0.0)
      rake (< 13.0)
    cliver (0.3.2)
    contracts (0.16.0)
    cucumber (3.1.2)
      builder (>= 2.1.2)
      cucumber-core (~> 3.2.0)
      cucumber-expressions (~> 6.0.1)
      cucumber-wire (~> 0.0.1)
      diff-lcs (~> 1.3)
      gherkin (~> 5.1.0)
      multi_json (>= 1.7.5, < 2.0)
      multi_test (>= 0.1.2)
    cucumber-core (3.2.1)
      backports (>= 3.8.0)
      cucumber-tag_expressions (~> 1.1.0)
      gherkin (~> 5.0)
    cucumber-expressions (6.0.1)
    cucumber-tag_expressions (1.1.1)
    cucumber-wire (0.0.1)
    diff-lcs (1.3)
    docile (1.3.2)
    ffi (1.11.1)
    gherkin (5.1.0)
    mini_mime (1.0.2)
    mini_portile2 (2.4.0)
    multi_json (1.14.1)
    multi_test (0.1.2)
    nokogiri (1.10.4)
      mini_portile2 (~> 2.4.0)
    parallel (1.17.0)
    parser (2.6.4.1)
      ast (~> 2.4.0)
    phantomjs (2.1.1.0)
    poltergeist (1.18.1)
      capybara (>= 2.1, < 4)
      cliver (~> 0.3.1)
      websocket-driver (>= 0.2.0)
    power_assert (1.1.5)
    powerpack (0.1.2)
    public_suffix (4.0.1)
    rack (2.0.7)
    rack-test (1.1.0)
      rack (>= 1.0, < 3)
    rainbow (2.2.2)
      rake
    rake (12.3.3)
    rspec (3.8.0)
      rspec-core (~> 3.8.0)
      rspec-expectations (~> 3.8.0)
      rspec-mocks (~> 3.8.0)
    rspec-core (3.8.2)
      rspec-support (~> 3.8.0)
    rspec-expectations (3.8.4)
      diff-lcs (>= 1.2.0, < 2.0)
      rspec-support (~> 3.8.0)
    rspec-mocks (3.8.1)
      diff-lcs (>= 1.2.0, < 2.0)
      rspec-support (~> 3.8.0)
    rspec-support (3.8.2)
    rubocop (0.49.1)
      parallel (~> 1.10)
      parser (>= 2.3.3.1, < 3.0)
      powerpack (~> 0.1)
      rainbow (>= 1.99.1, < 3.0)
      ruby-progressbar (~> 1.7)
      unicode-display_width (~> 1.0, >= 1.0.1)
    ruby-progressbar (1.10.1)
    simplecov-html (0.10.2)
    test-unit (3.3.3)
      power_assert
    thor (0.20.3)
    unicode-display_width (1.6.0)
    websocket-driver (0.7.1)
      websocket-extensions (>= 0.1.0)
    websocket-extensions (0.1.4)
    xpath (3.2.0)
      nokogiri (~> 1.8)

PLATFORMS
  ruby

DEPENDENCIES
  aruba (~> 0.14)
  benchmark-ips
  bundler
  capybara (< 3)
  cucumber (~> 3.1)
  phantomjs
  poltergeist
  rake (~> 12.0)
  rspec (~> 3.2)
  rubocop (= 0.49.1)
  simplecov!
  test-unit

BUNDLED WITH
   1.17.3
```

</details>

adam12: Thanks! I'm not sure if this is the issue but it hopefully eliminates version mismatch.



> All this makes me thing of José Valim never understanding/liking that libraries in ruby didn't use the `.lock` file to avoid problems like this.

I think a lot of people who have history in Ruby took what Yehuda said way back about not shipping `.lock` files with libraries and never updated their knowledge of it. I know [I was guilty of this](https://github.com/seanhandley/fraudrecord/pull/5) even not too long ago. That said, [best practice seems to have changed](https://bundler.io/man/bundle-install.1.html#THE-GEMFILE-LOCK) and now shipping with `.lock` files is encouraged. 

I'll try out this lock file and see if it gets me any further. I'd be open to submitting a PR to fix up the Capybara driver but I suspect that might lead to another yak shave, which might be dropping Travis.

Will advise.

adam12: Unfortunately that didn't get me any further. Maybe it's an issue with me using FreeBSD. I still had to swap out the Capybara driver but this looks like an issue with the stdout generated, and not the test output HTML.

Here's [the output from Cucumber](https://gist.github.com/adam12/600f0eee7158a5713034bbace73fa6aa) in case you might have a hint on what's going on, but I'll see if I can run it on Linux to identify FreeBSD as the culprit. Then work backwards I guess.

PragTob: From my reading it expects it to say that an rspec coverage report has been generated:

```
-Coverage report generated for RSpec
```

But the output doesn't say RSpec:

```
Coverage report generated for Unit Tests
```

adam12: I noticed that. I hadn't made any changes to the project outside of Capybara, so I had thought it was a dependency issue.

The fake project has a Gemfile too without a lock. I wonder if it's an issue with the fake project?

Mysteries. 🕵

PragTob: It sounds like test framework detection is botched... I dunno hard to say :(

PragTob:
rubocop is complaining: spec/faked_project/minitest/test_helper.rb:8:3: C: Style/StderrPuts: Use warn instead of $stderr.puts to allow such output to be disabled.

  $stderr.puts "No SimpleCov config file found!"

I might get to fixing this up myself tomorrow or so but if you get to it before I do, ofc feel free!

adam12:
I have a fix locally but left it broken intentionally on remote as I’m still not 100% sure my fix works in all scenarios and didn’t want a premature merge. 

Leave it with me :) going to get to the bottom of it eventually.

spec/defaults_spec.rb
adam12: Just FYI, I was able to resolve this by forcing bundler to install in a local path and not install globally through `GEM_HOME`. Not really sure why but I suspect it might have to do with the nesting of dependencies somehow?

But `bundle install --path some_path` just made it work out of the box. 🕵

adam12:
I'm still exploring if this solution is right in all cases. I had to re-work the original patch when writing the feature specs, but the original patch indeed worked locally when I used it on a project.

I suspect that there's an issue with load order; when Minitest is loaded, the hook can be set correctly and everything functions well. If Minitest isn't loaded, the original `at_exit` mechanism is used, but is then negated by Minitest loading afterwards.

A Minitest plugin might work better here, but that still needs to be explored.

PragTob:
@adam12 thanks a lot for all the work you put into this! `at_exit` handling seems really gnarly in this case :) Let me know when this is ready so we can get it merged!

PragTob:
All the tests except for the newest ruby version are failing with a slightly different coverage... which is an interesting edge case :thinking: 

Not sure what to do about it though... making it conditional, a range, trying to find a bug (I suspect it's in ruby itself...) - thougs @adam12 ?

Also, thanks a lot!

adam12:
Yeah, I'm not too sure what to make of it right now. I've been pondering it all weekend.

I'll probably have to establish which value I believe is correct and then work backwards from there. I'm guessing it's from changes I made because CI for master tests those older versions with correct coverage.

I'll keep poking at it.

PragTob:
@adam12 the failure is from your newly introduced feature test (well done for introducing it and thank you! :green_heart: ) - so in that sense it is from changes you made, but the actually measurement might always have been broken :) I'd suggest looking at the HTML report and see what's correct

PragTob:
@adam12 Hey, how's it going? :) Do you wanna keep poking at it or should I see if I could/should take over? It'd be no bother for me, you already did a lot :green_heart: If you wanna keep poking at it I also understand and will do things elsewhere.

I _really_ want to have this as I believe it'll fix quite some issues for people :rocket:

adam12:
Hello @PragTob!

I haven't forgotten about it but my free time ebbs and flows. Hoping to look at it again this week.

If you're interested in taking a poke at the coverage values and seeing if you can spot the discrepancy I'd definitely not oppose :)

Cheers.

PragTob:
Hi @adam12 thank you!

I'll first stick with the branch coverage so we can get that relased as a preview and then if you haven't fixed it I'll take a poke at it. Thanks so much! :ok_hand:

adam12:
> I'll first stick with the branch coverage so we can get that relased as a preview and then if you haven't fixed it I'll take a poke at it. Thanks so much!

Looks like branch coverage might be our issue here as well?

image

Also, I'll be rebasing against master unless you've started working on this branch? LMK.

PragTob:
@adam12 nope haven't started working on it. Rebase away!

Na that's not branch coverage/branch coverage doesn't seem to be enabled. However, the number of relevant lines seems to have decreased and somewhat rightfully so. I wouldn't think `begin` is relevant as there isn't really any expression there.

PragTob:
#855 is there now trying to get this in.

PragTob:
And #855 appears to be ready. Closing this in favor of it.

PR #757, opened 2019-11-11
Update references to Minitest constant

adam12:
The camel-case constant `MiniTest` will be dropped in Minitest 6. It's been `Minitest` for a number of years now so this change should be fairly safe.

### DIFF
--- a/lib/simplecov/command_guesser.rb
+++ b/lib/simplecov/command_guesser.rb
@@ -48,6 +48,8 @@ module SimpleCov
           "RSpec"
         elsif defined?(Test::Unit)
           "Unit Tests"
+        elsif defined?(Minitest)
+          "Minitest"
         elsif defined?(MiniTest)
           "MiniTest"
         else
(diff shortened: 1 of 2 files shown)


lib/simplecov/command_guesser.rb
```diff
@@ -48,8 +48,8 @@ def from_defined_constants
           "RSpec"
         elsif defined?(Test::Unit)
           "Unit Tests"
-        elsif defined?(MiniTest)
-          "MiniTest"
+        elsif defined?(Minitest)
```
bf4: why not both?

adam12: I'd be OK with both, but since the original constant was going to be completely removed, and the new one has been current for _ages_, it just seemed like debt.

But if we want it back I don't see the harm :)

PragTob: @adam12 yes please add it back :) It's a small price to pay to maybe still support some people with very old versions.

adam12: I'll do just that. Thanks!

adam12:
Changes made. :+1:

PR #767, opened 2019-11-27
Rubocop improvements

deivid-rodriguez:
#754 introduced several problems in the rubocop setup:

* It updated the configuration to use `Style/TrailingCommaInArrayLiteral` and `Style/TrailingCommaInHashLiteral`. However, those cops [were introduced in rubocop 0.53.0](https://rubocop.readthedocs.io/en/latest/cops_style/#styletrailingcommainarrayliteral), and we're using rubocop 0.49.0.

* The code was changed to enforced exactly the opposite style being configured. The cops were setup to enforce a trailing comma at the end of hash and array literals, but the code was updated to remove that trailing comma.

The reason why the rubocop run is current passing is because the cops are not recognized in rubocop 0.49.0, and get ignored.

In this PR, I update the configuration of these cops to match the current style of the code, and also upgrade rubocop to 0.53.0 so that the cops get applied. Also, I'm fixing/ignoring the new offenses found by the new rubocop version.

### DIFF
--- a/lib/simplecov/formatter/simple_formatter.rb
+++ b/lib/simplecov/formatter/simple_formatter.rb
@@ -8,7 +8,7 @@ module SimpleCov
     class SimpleFormatter
       # Takes a SimpleCov::Result and generates a string out of it
       def format(result)
-        output = "".dup
+        output = +""
         result.groups.each do |name, files|
           output << "Group: #{name}\n"
           output << "=" * 40
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -103,7 +103,7 @@ Style/StringLiterals:
   Description: Allow double-quoted strings without interpolation.
   EnforcedStyle: double_quotes
 
-Style/FileName:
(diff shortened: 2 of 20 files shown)


lib/simplecov/formatter/simple_formatter.rb
```diff
@@ -8,7 +8,7 @@ module Formatter
     class SimpleFormatter
       # Takes a SimpleCov::Result and generates a string out of it
       def format(result)
-        output = "".dup
+        output = +""
```
PragTob: the.... what? I think I need to look this one up...

PragTob: Well... rubocop always teaching me new things. https://www.rubydoc.info/gems/rubocop/0.61.1/RuboCop/Cop/Performance/UnfreezeString

deivid-rodriguez:
Thanks @PragTob!

PR #768, opened 2019-11-29
Bump CI things

deivid-rodriguez:
This PR bumps some software used in TravisCI, such as rubygems and the ruby versions we test against.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: ruby
 
 before_install:
-  - gem update --system '3.0.4'
+  - gem update --system '3.0.6'
   - gem i bundler -v '2.0.2'
 
 bundler_args: --without benchmark --jobs=3 --retry=3
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: ruby
 
 before_install:
-  - gem update --system '3.0.6'
   - gem i bundler -v '2.0.2'
 
 bundler_args: --without benchmark --jobs=3 --retry=3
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,16 +1,16 @@
 language: ruby
 
 before_install:
-  - gem i rubygems-update -v '3.0.4' && update_rubygems
+  - gem update --system '3.0.6'
```
PragTob: Do you really think we need this? I think we bumped/froze it once due to some problem. I'd hope that our recently released versions come with a good enough rubygems version. So, maybe we can drop this line entirely?

.travis.yml
```diff
@@ -1,16 +1,16 @@
...
 bundler_args: --without benchmark --jobs=3 --retry=3
 
 matrix:
   include:
-    - rvm: 2.4.6
-    - rvm: 2.5.5
-    - rvm: 2.6.3
+    - rvm: 2.4.9
+    - rvm: 2.5.7
+    - rvm: 2.6.5
```
PragTob: :+1: :ok_hand:

.travis.yml
deivid-rodriguez: I can try. Normally you do this when you need bug fixes from the latest rubygems, but maybe it's not necessary here. Let me remove it.

PragTob: yeah I know that's why we did it, but I can't remember when we did it anymore so I thought right now might be as good a time as any to check if we need that :grin:

deivid-rodriguez: It's passing so seems good!

PR #775, opened 2019-12-03
Upgrade travis

PragTob:


PragTob:
Doesn't seem to know the new ruby preview and JRuby is also failing with weird errors...

PR #681, opened 2018-05-25
WIP: easier manual collation of result sets

ticky:
This adds a new entry point for collating multiple `.resultset.json` files, when generated by parallelised, multi-machine test runs (for example when used with [Knapsack Pro](https://www.knapsackpro.com) on Circle, Travis or Buildkite).

In these environments, the current behaviour of merging with the “last” result set is defeated by the fact that runs occur on isolated containers or machines which don’t know about the “last” run.

For these cases, you can create, for instance, a Rakefile or script file, which calls `SimpleCov.collate` with a list of paths to `.resultset.json` files from each parallel test run. In our case, we tar these and upload them as artifacts.

```ruby
# coverage_report.rake
namespace :coverage do
  task :report do
    require 'simplecov'

    SimpleCov.collate Dir["simplecov-resultset-*/.resultset.json"], 'rails' do
      formatter SimpleCov::Formatter::MultiFormatter.new([
        SimpleCov::Formatter::SimpleFormatter,
        SimpleCov::Formatter::HTMLFormatter
      ])
    end
  end
end
```

Calling `SimpleCov.collate` takes the same type of configuration block as `SimpleCov.start` but will _not_ collect coverage stats itself, only combine the results from the input JSON files, and configure `SimpleCov` to format the results when the task exits in the same way as if you called `SimpleCov.start`.

To Do:
- [x] Initial pass at feature
- [x] Test in-place in a project
- [ ] Add specs
- [x] Add documentation

PR #780, opened 2019-12-04
Allow manual merging

deivid-rodriguez:
This my WIP version of #681. It would probably be better to keep the discussion in that PR, but to do that I would need permissions to push to @ticky's fork.


lib/simplecov.rb
```diff
@@ -54,6 +54,40 @@ def start(profile = nil, &block)
...
+
+      results = result_filenames.map { |filename| JSON.parse(File.read(filename)) || {} }.reduce(&:merge).map do |command_name, data|
+        # Re-create each included instance of SimpleCov::Result from the stored run data.
+        SimpleCov::Result.from_hash(command_name => data)
+      end
+
+      # Use the ResultMerger to produce a single, merged result, ready to use.
+      @result = SimpleCov::ResultMerger.merge_results(*results)
+      SimpleCov::ResultMerger.store_result(@result) if @result
+      @result
```
deivid-rodriguez: I needed to add the above two lines to get tests passing because I noticed that the merged coverage resultset was not being written to disk. That surprised me though because I thought some people were using this, but it seems to me that without these lines this feature can't work (as the cucumber scenario was showing). Maybe I'm missing something? :thinking:

deivid-rodriguez:
This should be ready now, I think?

I rebased the old PR and investigated and fixed the cucumber failures (with the help of https://github.com/cucumber/aruba/pull/675). I also address some not yet addressed comments about the documentation from the previous PR.

deivid-rodriguez:
Well, I'm now trying this in my app and it's not quite working as I expect. I need to have a closer look, so this is back to WIP.

lib/simplecov.rb
PragTob: Right now your guess is as good as mine I haven't read that code in a long while and need to fresh up on it again :sweat_smile:

PragTob:
Thanks so much @deivid-rodriguez (and @ticky !) for this :green_heart: :tada: :rocket: 

As it's WIP atm I haven't taken a look yet, will take one at another time but just wanted to express my gratitude :hugs: 

IMG_20181116_075339_01

lib/simplecov.rb
deivid-rodriguez: I _think_ the reason is that it's written to disk in the `at_exit` hook, which the tests doesn't really run. But I need to dig in deeper.

deivid-rodriguez:
Thanks for your encouraging and positive attitude :) I'll try to get this over the line!

lib/simplecov.rb
ticky: I think this was an oversight on my part - we're relying on plugins' output, which are written regardless of the merged result set JSON being stored. Making it write this makes sense to me. Good catch!

ticky:
Thanks for pushing ahead with this @deivid-rodriguez! Your changes are looking good to me - I’ll try them out in place of my branch and see if things still work as expected.

deivid-rodriguez:
Ok, I think I fixed the issue. Previously results were being incorrectly merged using `Hash#merge`, and the `ResultMerger.merge_results` was being passed a single incorrectly merged result. Essentially the previous version was merging resultsets correctly at the file level, but not at the line level. See https://github.com/colszowka/simplecov/pull/780/commits/4fedabdf070a7b02dd0872b3df3717ad4a4057ec.

This now works as I expect against my library, I'll look into getting the last fix covered.

deivid-rodriguez:
@ticky Make sure you use the latest commit too. I think your different runners are correctly merging the set of files that the cover all together, but you might not be properly getting the set of lines covered for each file properly merged. So, I would expect the last commit to increase the total coverage in your application.

ticky:
All working for us here on 4fedabdf070a7b02dd0872b3df3717ad4a4057ec.

Looks like a minor change in coverage data reporting, but it seems about as expected.

Nice work @deivid-rodriguez :)

PragTob:
@ticky thanks for testing this out! :green_heart: 

Really good work @deivid-rodriguez ! I have a (minor) knee surgery tomorrow, not sure if I get to look at this properly before ~Friday/Saturday so please excuse my tardyness but I really want to get this in :)

IMG_20181111_110737

deivid-rodriguez:
@PragTob Hope your surgery goes well!

Regarding this PR, I still want to give it another look and try a few extra things, so no rush. Will let you know when I consider it ready for review!

deivid-rodriguez:
Found one more issue where only the first command_name => coverage_data pair on each result set was getting merged. Fixed by https://github.com/colszowka/simplecov/pull/780/commits/4d3492d99b0ed67af94c4762ff0e592bf823a142.

README.md
```diff
@@ -415,12 +415,81 @@ end
...
+be taken into account when building the coverage report.
+
+#### Timeout for merge
+
+Of course, your cached coverage data is likely to become invalid at some point. Thus, when automatically merging
+subsequent test runs, result sets that are older than `SimpleCov.merge_timeout` will not be used any more. By default,
+the timeout is 600 seconds (10 minutes), and you can raise (or lower) it by specifying `SimpleCov.merge_timeout 3600`
+(1 hour), or, inside a configure/start block, with just `merge_timeout 3600`.
+
+You can deactivate this automatic merging altogether with `SimpleCov.use_merging false`.
```
PragTob: :tada: Thanks for the improved documentation! :ok_hand:

README.md
```diff
@@ -415,12 +415,81 @@ end
...
 You normally want to have your coverage analyzed across ALL of your test suites, right?
 
-Simplecov automatically caches coverage results in your (coverage_path)/.resultset.json. Those results will then
-be automatically merged when generating the result, so when coverage is set up properly for Cucumber and your
-unit / functional / integration tests, all of those test suites will be taken into account when building the
-coverage report.
+Simplecov automatically caches coverage results in your (coverage_path)/.resultset.json.
 
-There are two things to note here though:
+### Between test runs under the same environment
```
PragTob: I think _environmnet_ is a bit ambiguous as it might refer to dev/test/prod (yes it should usually be test) - maybe machine/execution environment or something else?

README.md
```diff
@@ -415,12 +415,81 @@ end
...
+This means you can configure a separate formatter for the collated output. For instance, you can make the formatter in `SimpleCov.start`
+the `SimpleCov::Formatter::SimpleFormatter`, and only use more complex formatters in the final `SimpleCov.collate` run.
+
+```ruby
+# spec/spec_helper.rb
+require 'simplecov'
+
+SimpleCov.start 'rails' do
+  # Disambiguates individual test runs
+  command_name "Job #{ENV["TEST_ENV_NUMBER"]}" if ENV["TEST_ENV_NUMBER"]
```
PragTob: As we use `command_name` here I'm not sure if maybe this section should come after the _test suite name_ section or should at least reference it

README.md
```diff
@@ -415,12 +415,81 @@ end
...
+namespace :coverage do
+  task :report do
+    require 'simplecov'
+
+    SimpleCov.collate Dir["simplecov-resultset-*/.resultset.json"], 'rails'
+  end
+end
+```
+
+`SimpleCov.collate` also takes a block for configuration, just the same as `SimpleCov.start` or `SimpleCov.configure`.
```
PragTob: In the example above you seem to be passing a profile name to `collate` - that isn't yet documented in here :)

features/test_unit_collate.feature
```diff
@@ -0,0 +1,43 @@
...
+      | lib/faked_project/some_class.rb         | 80.0 %   |
+      | lib/faked_project/framework_specific.rb | 75.0 %   |
+      | lib/faked_project/meta_magic.rb         | 100.0 %  |
+      | test/meta_magic_test.rb                 | 100.0 %  |
+      | test/some_class_test.rb                 | 100.0 %  |
+
+      # Note: faked_test.rb is not appearing here since that's the first unit test file
+      # loaded by Rake, and only there test_helper is required, which then loads simplecov
+      # and triggers tracking of all other loaded files! Solution for this would be to
+      # configure simplecov in this first test instead of test_helper.
```
PragTob: huh :thinking: 

I'd like the test setup in our repo to be correct :sweat_smile: Might be out of scope here, as you used what existed. Wondering if configuring simplecov in the test file is really the only thing we can do - couldn't we just load test_helper before?

features/test_unit_collate.feature
```diff
@@ -0,0 +1,43 @@
...
+      | All Files | 91.38%   | 6     |
+
+    And I should see the source files:
+      | name                                    | coverage |
+      | lib/faked_project.rb                    | 100.0 %  |
+      | lib/faked_project/some_class.rb         | 80.0 %   |
+      | lib/faked_project/framework_specific.rb | 75.0 %   |
+      | lib/faked_project/meta_magic.rb         | 100.0 %  |
+      | test/meta_magic_test.rb                 | 100.0 %  |
+      | test/some_class_test.rb                 | 100.0 %  |
```
PragTob: Thanks for putting all the time into an integration test :ok_hand:

lib/simplecov.rb
```diff
@@ -54,6 +54,42 @@ def start(profile = nil, &block)
...
+    #    OR
+    #   SimpleCov.collate Dir["simplecov-resultset-*/.resultset.json"] do
+    #     add_filter 'test'
+    #   end
+    #    OR
+    #   SimpleCov.collate Dir["simplecov-resultset-*/.resultset.json"], 'rails' do
+    #     add_filter 'test'
+    #   end
+    #
+    # Please check out the RDoc for SimpleCov::Configuration to find about available config options
```
PragTob: Might wanna mention README for a more in depth description of what collate does

lib/simplecov.rb
```diff
@@ -54,6 +54,42 @@ def start(profile = nil, &block)
...
+      configure(&block) if block_given?
+      self.running = true
+      self.pid = Process.pid
+
+      results = result_filenames.map do |filename|
+        # Re-create each included instance of SimpleCov::Result from the stored run data.
+        (JSON.parse(File.read(filename)) || {}).map do |command_name, coverage|
+          SimpleCov::Result.from_hash(command_name => coverage)
+        end
+      end.flatten
```
PragTob: Shouldn't we be able to use `flat_map` ?

lib/simplecov.rb
```diff
@@ -54,6 +54,42 @@ def start(profile = nil, &block)
...
+    #     add_filter 'test'
+    #   end
+    #    OR
+    #   SimpleCov.collate Dir["simplecov-resultset-*/.resultset.json"], 'rails' do
+    #     add_filter 'test'
+    #   end
+    #
+    # Please check out the RDoc for SimpleCov::Configuration to find about available config options
+    #
+    def collate(result_filenames, profile = nil, &block)
```
PragTob: I like that we have a nice feature test covering this, I'm just wondering that for a rather complex feature (although it looks fairly simple in its implementation) if a unit test or 2 wouldn't make sense.

I was thinking like:
1. a simple merge
2. there's nothing to merge (just one file)
3. no files are matching


 What do you think?

PragTob:
The failing test is the JRuby flaky one I gotta take care off... chose to not rerun it as I figured you still wanted to update it here

deivid-rodriguez:
Thanks for the review @PragTob!

I still want to fix some issues with the current implementation. For example, a correct merged report is now generated _twice_, one inline, and one in the `at_exit` hook, which is confusing at least. Then I wanted to have a look at test coverage and see how I can improve it. Now I'll also add "Address @PragTob's feedback" to the list :)

PragTob:
Sounds like a plan! :ok_hand: 

IMG_20191011_074652

features/test_unit_collate.feature
PragTob: I just realized that this is like this in other tests as well... please ignore this comment then :)

README.md
deivid-rodriguez: Fixed by https://github.com/colszowka/simplecov/pull/780/commits/90d67d1c316273d69fef45db18ddf66fd6c425c0.

README.md
deivid-rodriguez: Fixed by https://github.com/colszowka/simplecov/pull/780/commits/ca6635b679421ef884d30769fbd6e95745fa3695.

README.md
deivid-rodriguez: Fixed by https://github.com/colszowka/simplecov/pull/780/commits/3ac94a1c255e1da0d1598af821d0838f810acc06.

features/test_unit_collate.feature
deivid-rodriguez: Marking as resolved since it was a "false positive" :)

features/test_unit_collate.feature
deivid-rodriguez: Thanks to @ticky :purple_heart:!

lib/simplecov.rb
deivid-rodriguez: Fixed by https://github.com/colszowka/simplecov/pull/780/commits/99c0495ac8fcb5d0a268dfc4f638b4c086c6c97e.

lib/simplecov.rb
deivid-rodriguez: Fixed by https://github.com/colszowka/simplecov/pull/780/commits/e5600ffcb43bbd915e9839ae18b210aacd6ae209.

deivid-rodriguez:
So I created some unit tests for this, but I run into several issues while creating them that I think might be related to #820. So I think I might've fixed that issue?

What I did was essentially making sure that when reading reports from disk into a hash, or manipulating them, all keys are always symbols. Do we have an regression scenario for #802? If so, I can extract the relevant changes from this PR into a separate PR, and add that regression test in there.

PragTob:
@deivid-rodriguez is the last ticket reference to #802 a mistake and you wanted to do #820 ? If so then we almost have one, I started writing one (generated an old resultset.json, created a project for it etc) but then got sick and took a break :grin: If I don't feel worse tomorrow I'll likely finish it up tomorrow and will check against your branch. Although, imo it's likely that this isn't #820 as I think #820 has to do with the different file structure (extra level) not the keys being symbols. 

On a slightly related note I was pondering removing the "symbolization" of keys as it seems to be a waste of memory & run time while we could us them without them being symbolized but I wanted to double check and see how far the symbolization goes.

deivid-rodriguez:
Yes, I corrected the mistake now. I'm not sure it wasn't #820, but since I ended up removing the `symbolize_names_of_coverage_results` which is what's crashing there, I figure this issue might be gone.

Regarding removing the symbolization, I think that's a good move, yeah. I was about to do that to fix the issue I was having (some "file name" keys were symbols, and others string, so they wouln't merge fine). But I ended up symbolizing everywhere instead, since it looked simpler to achieve.

deivid-rodriguez:
Oh, I also made this change: https://github.com/colszowka/simplecov/pull/780/commits/93c771d3758f8ba31382de84b96277bc873ee1bc, because `branches: nil` was sneaking into merged reports, even if none of the merged reports had any branch information in them. Hopefully it's correct?

PragTob:
Yeah it crashed in the symbolization, but I think it'll then just crash somewhere else where it tries to traverse the tree - unless the time is old (meaning time in the resultset is after the timeout) in which case it wouldn't do that and I'd have to adjust my test :D 

Not sure about the other commit, I'd have to check impact in other places. I think it's more in line with what normal Coverage returns so that's good - sometimes I prefer to have all keys but just have them be empty but... tradeoffs :man_shrugging:

PragTob:
@deivid-rodriguez so me trying to reproduce the failure is at the branch `fix-crash-against-old-resultset-json` but.. it passes. So it doesn't seem to be what I assumed it was (incompatibility with the old resultset.json) and it just runs against it and works. Maybe it's just not triggered or whatever the condition is isn't quite right in the test setup.

PragTob:
@deivid-rodriguez  Okay I double checked with #824 and it first it seemed that it solved the crash, but it then highlighted a couple of errors in my test setup that I need to fix up (the paths in the saved coverage.json need to be the same ones as the local machine and I need to update the time to not go beyond the merge timeout as your changes move the crash to that time). Thanks! :green_heart:

lib/simplecov.rb
PragTob: We have those now! :clap:

lib/simplecov.rb
```diff
@@ -46,15 +46,52 @@ class << self
...
+    #     add_filter 'test'
+    #   end
+    #
+    # Please check out the RDoc for SimpleCov::Configuration to find about
+    # available config options, or checkout the README for more in-depth
+    # information about coverage collation
+    #
+    def collate(result_filenames, profile = nil, &block)
+      if result_filenames.empty?
+        $stderr.printf("There's no reports to be merged\n")
```
PragTob: Maybe we'd rather wanna raise here? In that case I'd also probably convert it to a guard clause. What do you think?

spec/simplecov_spec.rb
```diff
@@ -230,10 +230,95 @@
...
+        collated = JSON.parse(File.read(resultset_path), symbolize_names: true).transform_values { |v| v.reject { |k| k == :timestamp } }
+        expect(collated).to eq(expected)
+      end
+    end
+
+    context "when multiple reports to be merged" do
+      before do
+        @result1 = SimpleCov::Result.new(resultset1)
+        @result1.command_name = "result1"
+        SimpleCov::ResultMerger.store_result(@result1)
```
PragTob: The way I see it those could all be local variables, right? (same for similar variables here)

spec/simplecov_spec.rb
```diff
@@ -230,10 +230,95 @@
...
+        FileUtils.mv resultset_path, "#{resultset_path}.final"
+
+        @result2 = SimpleCov::Result.new(resultset2)
+        @result2.command_name = "result2"
+        SimpleCov::ResultMerger.store_result(@result2)
+        FileUtils.mv resultset_path, "#{resultset_path}2.final"
```
PragTob: The code in these before blocks (for resutl1/2 individually) seems repeated enough to me that I'd consider extracting a method like "create_result" or something to help with those and have that centralized. If you agree, then please extract it. If not, then I'm also fine leaving them where they are :)

spec/simplecov_spec.rb
```diff
@@ -230,10 +230,95 @@
...
+      end
+
+      after do
+        SimpleCov.clear_result
+        FileUtils.rm resultset_path
+        FileUtils.rm "#{resultset_path}.lock"
+        FileUtils.rm "#{resultset_path}.final"
```
PragTob: Same-ish thing for the the after block smells like it might be better as a small helper method here in the spec

deivid-rodriguez:
@PragTob I have no problem with this getting merged now, although it contains a bunch of commits which seem completely unrelated to the coverage merging (although I did add them because of issues when getting the unit specs to pass, so I guess there's some relation after all).

So, your choice, I don't mind extracting those commits to a separate PR, and I don't care about rebasing stuff and fixing merge conflicts. Just let me know.

Regarding your feedback, yeah, I just wanted to push bare passing specs but I was planning to refactor and improve them afterwards.

I'll get this ready for you later today hopefully. I'd also like to do some commit history cleanup since there's a lot of stuff in here that can be squashed together, but I can also skip doing that if you prefer the original verbose git history of this PR.

PragTob:
@deivid-rodriguez Thanks :green_heart: All volunteers here so no worries, but if you managed to do it by this evening that'd be super great. I'll leave my work on #824 alone until it's done or tomorrow morning or so then.

I'd appreciate a bit of squashing (not strictly necessary from my side but some fewer seems like a nice idea) :) While I sympathize with extracting it to a separate PR I think it's fine to have in here - these things sometimes escalate and happen but imo it's small enough (and I've already reviewed it anyhow) to keep in here imo :)

Thanks heaps!

IMG_20180805_081900

lib/simplecov.rb
deivid-rodriguez: Addressed by https://github.com/colszowka/simplecov/pull/780/commits/2f01dff18ace47f393988eec5680249b995d4970.

spec/simplecov_spec.rb
deivid-rodriguez: Addressed by https://github.com/colszowka/simplecov/pull/780/commits/c5a2a623199e3130fd5f4a0fdfc95ee818824161.

spec/simplecov_spec.rb
deivid-rodriguez: Addressed by https://github.com/colszowka/simplecov/pull/780/commits/f001222534850f1f5991243b7c2d5f5663e3e73b.

spec/simplecov_spec.rb
deivid-rodriguez: Addressed by https://github.com/colszowka/simplecov/pull/780/commits/6de9ebf3df3eeb39b4bf16f2403ad98c9d3148e2.

deivid-rodriguez:
I addressed your feedback. If you're good with the current version I can do a pass of squashing.

PragTob:
@deivid-rodriguez I'm good with it, ruby 2.4.9 seems to have a legit failure though :scream: 

Anyhow, thanks a bunch for the work! :green_heart:

deivid-rodriguez:
It seems like a flaky... :(

deivid-rodriguez:
I can reproduce locally using the same test seed, will have a look.

deivid-rodriguez:
It should be ok now. I'll do the squashing.

deivid-rodriguez:
I think I detected a potential bug while squashing, I'll have a closer look later.

deivid-rodriguez:
m
Ok, so I pushed a cleaner git history, which consists of the following:

* 8977f97 and 8f2f2a3 implement essentially the same feature as https://github.com/colszowka/simplecov/pull/731, so they essentially supersedes that PR. You can now call `SimpleCov.run_exit_tasks!` directly, and exit tasks will be avoided in that case. I added this feature because I needed it to get cucumber scenarios passing, since some of the magic was happening in the `at_exit` hook making it really difficult to test. I kept @macumber's attribution since I used his idea and part of the code.

* 0234c60, 0c33f1a and 5061173 implement "full symbolization", which might or might not help with #820, since it removes the code path where the crash happens. I needed this because my unit tests for merging reports were getting confused because some reports had symbol keys and others string keys.

* 3409893 as explained before removes the ":branches" section from reports if branch coverage is not enabled.

* f02b0a4 and 8ff1295 fix some test suite state leaks causing order dependent failures.

* 885ddd5 fixes a typo.

* b64b148 improves our test app scenario setup so that it doesn't depend on scenarios setting up LOAD_PATHs.

* 863bf38 implements the result collation feature.

deivid-rodriguez:
That "potential bug" was a false alarm by the way :sweat_smile:.

deivid-rodriguez:
It's a flaky world! I'll have a look at the new failure :)

PragTob:
Yay flakies! :dancer: :roll_eyes: 

Good luck with those. I guess instead of waiting I might just rebase my branch against this one and work from there, might be the most efficient so close to the finish line.

Also btw you're probably better equipped than me to write the Changelog (doesn't need to be within this PR, can be separate) 

Full symbolization helps with #820 but doesn't fully fix it (it helps for the case where the report is outdated) - fixing it will sadly be somewhat "dirty" or at least that's the only way I know how to (rescuing errors from the code that does the actual merging, print a warning, return empty value - while not great this helps us to not repeat this once we change the format again)

Cheers and a bunch of thanks!

deivid-rodriguez:
I fixed the flaky and wrote changelog entries!

ticky:
Thanks so much for your work getting this over the line! Super excited for this to hit a release :)

PragTob:
@ticky it's live now as 0.18.0.beta2 - if you could give it a try that'd be highly appreciated! :)

ticky:
Working great for us! Loving it :)

PragTob:
Happy to hear it!

WhatsApp Image 2019-06-26 at 08 42 20

PR #785, opened 2019-12-09
Add project metadata to the gemspec

orien:
Add `bug_tracker_uri`, `changelog_uri`, `documentation_uri`, and `source_code_uri` to the gemspec metadata.

These [project metadata](https://guides.rubygems.org/specification-reference/#metadata) will facilitate easy access to project information. The URI will be available on the [Rubygems project page](https://rubygems.org/gems/simplecov), via the rubygems API, and the `gem` and `bundle` command-line tools with the next release.


orien:
I think it's normal to point to Rubydoc site. It includes all the information in the readme and has class documentation to boot. It's easy for people to choose Github via the homepage or source URI if they so desire.

PragTob:
@orien seems fair, thank you! Makes me wanna brush up on our documentation though :sweat_smile: One thing after another though!

PR #787, opened 2019-12-09
Fix gemspec warning

deivid-rodriguez:
I noticed that running `gem build simplecov.gemspec` prints one warning:

```
WARNING:  description and summary are identical
WARNING:  See http://guides.rubygems.org/specification-reference/ for help
```

This PR fixes it.


PragTob:
Merged as 9bd26c0973a4e939f :tada:

Issue #781, opened 2019-12-05
Release Branch coverage


PragTob:
After merging #694 this issue is here to keep track of the functionality until initial release and helps me & others :)

Big thanks to @som4ik @tycooon @stepozer @klyonrad and many others for their valuable contributions!

* [x] Merge initial PR #694 
* [x] Merge corresponding html formatter PR colszowka/simplecov-html#64
  * [x] Address remaining code review feedback
* [x] Make sure problems with existing PR are solved ([ref1](https://github.com/colszowka/simplecov/pull/694#issuecomment-561092374) [ref2](https://github.com/colszowka/simplecov/pull/694#issuecomment-561095694))
  * [x] parallel tests support works as intended
  * [x] `:nocov:` blocks should be considered
* [x] Address [initial code review comments](https://github.com/colszowka/simplecov/pull/694#pullrequestreview-256178924)
* [x] Change [used option name](https://github.com/colszowka/simplecov/pull/694#discussion_r298918548) to use branch coverage/chanege coverage type
* [x] add some feature tests
  * [x] for instance with the neat [tester script](https://github.com/colszowka/simplecov/pull/694#issuecomment-562097006)
* [x] Investigate coverage reporting (it seems like it reports lines as covered when what we'd like to have are branches)
* [x] release pre release version for general/public testing (this and formatter)
* [ ] make it possible that the branch coverage percentage is displayed as the "main" coverage and all our checks take it instead of line coverage
* [x] also allow people to do coverage checks versus multiple metrics for instance I want my line coverage to be at least 95% and my branch coverage to be at least 90%.
* [x] figure out problems with the format data is stored in #781 
* [x] fix reported beta problems
* [x] release the "finished for now" branch coverage

PragTob:
:wave: 

Hi @som4ik - got around to writing a feature test for this. It seems to me (and my reading of the code) that we still report coverage on a per line basis - is this correct from your memory? :sweat_smile: 

What I mean by that is that we use branch coverage to figure out which lines are covered and then we mark those lines as covered and report them as such. I think what we want is that we report it on a per branch basis (so instead of 80/100 lines covered we wanna say 33/38 branches covered or something to the tune of this). 

Confirmation would be nice!

som4ik:
Hi @PragTob :wave: 
As I remember per line bases report is mandatory-default coverage report, passing `:branches` parameter to `Coverage.start` adds a new type of collected coverage result matrics, but doesn't cancel generating per line bases report.
[...]

PR #800, opened 2019-12-28
Nocov for branches + more branch test cases + simplifications

PragTob:
Started out as getting nocov support in for branches. Escalated from there to:

* [x] remove source file/branch support and inline them as there was no clear distinction what ended up in the module and what didn't, having it in one place will hopefully lead us to a better abstraction
* [x] extracted the Line and Branch classes to their own homes outside of the SourceFile class
* [x] see some code we should probably deal with (everything is treated as a branch while it seems there's always the top level condition and then it's branches) but tackling it without appropriate test coverage was scary so added a bunch of tests
* [x] do some refactoring (like the regex extraction of arguments) that hopefully won't escalate 
* [x] get the actual no cov comments skipping to work

#781


PragTob:
IMG_20190127_150119

PR #807, opened 2020-01-04
Make activation of branch coverage more sensible

PragTob:
Branch coverage is best to accompany line coverage as branch
coverage itself has some gaping holes in it (namely that it
doesn't care about code without conditionals).

coverage_criterion will most likely evolve to be the "primary
coverage criterion" for the full release.

This API right now should be future proof but also easier.

#781

### DIFF
--- a/README.md
+++ b/README.md
@@ -626,8 +626,9 @@ SimpleCov.refuse_coverage_drop
 Add branch coverage measurement statistics to your results
 
 ```ruby
+# or in configure or just SimpleCov.enable_coverage :branch
 SimpleCov.start do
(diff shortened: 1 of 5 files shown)


tycooon:
I think it would be nice to have ability to set the minimal acceptable coverage for each type. So that I can say that I want 100% line coverage and at least 95% branch coverage for example.

PragTob:
@tycooon that's a good point, at firest I thought we'd only want to set one "main" coverage type for which these would be executed  but you're right we'd probably want a mix of them as well. I'll update #781 accordingly. Thanks for chiming in! :green_heart:

PR #813, opened 2020-01-09
Only turn on requested coverage criteria on the Ruby Coverage side

PragTob:
when activating branch coverage before SimpleCov would also instruct Ruby to take Method coverage

#781

### DIFF
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -306,8 +306,8 @@ module SimpleCov
       :branch => :branches,
       :line => :lines
     }.freeze
-    def lookup_corresponding_ruby_coverage_name(cirterion)
-      CRITERION_TO_RUBY_COVERAGE.fetch(cirterion)
+    def lookup_corresponding_ruby_coverage_name(criterion)
+      CRITERION_TO_RUBY_COVERAGE.fetch(criterion)
     end
 
     #
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -52,7 +52,7 @@ module SimpleCov
       self.running = true
       self.pid = Process.pid
 
-      start_coverage_measurment
+      start_coverage_measurement
     end
 
     #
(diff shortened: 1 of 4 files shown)


lib/simplecov.rb
```diff
@@ -287,13 +287,29 @@ def start_coverage_measurment
...
+      end.to_h
+
+      Coverage.start(start_arguments)
+    end
+
+    CRITERION_TO_RUBY_COVERAGE = {
+      :branch => :branches,
+      :line => :lines
+    }.freeze
+    def lookup_corresponding_ruby_coverage_name(cirterion)
```
klyonrad: typo
```suggestion
    def lookup_corresponding_ruby_coverage_name(criterion)
```

lib/simplecov.rb
```diff
@@ -287,13 +287,29 @@ def start_coverage_measurment
...
+      CRITERION_TO_RUBY_COVERAGE.fetch(cirterion)
```
klyonrad:
```suggestion
      CRITERION_TO_RUBY_COVERAGE.fetch(criterion)
```

spec/simplecov_spec.rb
```diff
@@ -229,4 +229,27 @@
       end
     end
   end
+
+  # Normally wouldn't test private methods but just start has side effects that
+  # cause errors so for time this is pragmatic (tm)
+  describe ".start_coverage_measurement", :if => SimpleCov.coverage_start_arguments_supported? do
+    before :each do
+      # global state, yay!
```
klyonrad: cheeky comment, but it doesn't really explain why `@coverage_criteria` keeps being stored and needs to be cleared

PragTob: ah, I'll add some more then. It's stored in the eigenclass object of SimpleCov itself and that isn't restored between test runs. It's basically a Singleton object global.

IPart of me would like to change this, the other part of me knows that it might break too many things.

PragTob:
Thanks a lot for the feedback!

IMG_20190127_150119

PR #822, opened 2020-01-12
Add note how to use simplecov with knapsack_pro gem

ArturT:
I often get asked by users of my knapsack_pro gem how to make simplecov working with my tool. I noticed you have a section for `Notes on specific frameworks and test utilities` so I've created this PR. Let me know if this looks ok.

Here is preview:
https://github.com/ArturT/simplecov/tree/note-knapsack#notes-on-specific-frameworks-and-test-utilities

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,17 @@ to use SimpleCov with them. Here's an overview of the known ones:
   </tr>
   <tr>
     <th>
+      knapsack_pro
+    </th>
+    <td>
+      To make SimpleCov work with Knapsack Pro Queue Mode to split tests in parallel on CI jobs you need to [provide CI node index number to the `SimpleCov.command_name`](https://knapsackpro.com/faq/question/how-to-use-simplecov-in-queue-mode).
+    </td>
+    <td>
+      <a href="https://knapsackpro.com/faq/question/how-to-use-simplecov-in-queue-mode">Tip</a>
+    </td>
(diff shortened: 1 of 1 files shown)


ArturT:
Merging simplecov reports depends a lot on CI providers. Some CI servers would allow you to preview reports in web UI if you store results as assets. You can also use CodeClimate to convert simplecov report and send to CodeClimate. 
https://knapsackpro.com/faq/question/how-to-use-codeclimate-with-knapsack_pro

I'm wondering about creating some more general solution but it would require to store reports somewhere. Maybe a simple bash script to send reports temporarily to S3 and later merge by the last executed parallel job.

I've updated my [FAQ](https://knapsackpro.com/faq/question/how-to-use-simplecov-in-queue-mode) to also mention that you could use CodeClimate to merge reports.

@PragTob Thanks for merging PR.

PragTob:
@ArturT I wouldn't necessarily use code climate for the merging. Not sure how much merging of reports they do, plus reportedly at elast when running with parallel tests their results can be off from what we do which lies in the way they integrate with simplecov (through an internal undocumented file)

What you say is exactly what I'd recommend - send the intermediate files to some commonly accessible location and then let the last one/a final process merge them together. @ticky and @deivid-rodriguez have been working on a solution to include into simplecov for that, see: #780

Issue #815, opened 2020-01-11
Race Condition in parallel_tests integration: an outdated report may be written


PragTob:
Discovered/verified in #841

Basically apparently it can happen that we generate the correct report but override it with an outdated one. Discovered running the parallel tests part of the features introduced in #841 (currently tagged wip to avoid flakies, remove the wip tag if you want to test this out). It doesn't happen reliably. Might happen 3 times in a row or not at all for 20 runs (for me at least).

I gathered some examples here: https://github.com/colszowka/simplecov/pull/814#issuecomment-573313375

What you can see there is something like this:

```
Coverage report generated for (2/8) to /home/tobi/github/simplecov/tmp/aruba/project/coverage. 23 / 33 LOC (69.7%) covered.
Coverage report generated for (1/8), (2/8), (3/8) to /home/tobi/github/simplecov/tmp/aruba/project/coverage. 37 / 43 LOC (86.05%) covered.
Coverage report generated for (1/8), (2/8), (3/8), (4/8) to /home/tobi/github/simplecov/tmp/aruba/project/coverage. 42 / 47 LOC (89.36%) covered.
Coverage report generated for (1/8), (2/8) to /home/tobi/github/simplecov/tmp/aruba/project/coverage. 32 / 39 LOC (82.05%) covered.
```

Notice that the second too last has the right number of loc and coverage expected in the test and also seems to take everything into account (unsure why it says n/8 though, guess: I could run 8 in parallel but parallel tests only spawned 4 as there are only 4 test files) but is seemingly later overridden by an outdated report.

If possible we should of course find and fix that race condition ;)

nickshatilo:
+1 here

benissimo:
Does bumping up merge_timeout mitigate this? It looks like the number of tests involved is small enough that the default timeout of 10 minutes should be more than enough. Just curious whether that's a factor.

I ran into the following case with some long running test suites:

```
[osm.jruby] Using 'RSpecdevn3' as SimpleCov command name
[osm.jruby] Using 'RSpecdevn1' as SimpleCov command name
[osm.jruby] Using 'RSpecdevn4' as SimpleCov command name
[osm.jruby] Using 'RSpecdevn2' as SimpleCov command name
[osm.jruby] Coverage report generated for RSpecdevn3 to rails/coverage. 44377 / 168731 LOC (26.3%) covered.
[...]

PR #823, opened 2020-01-12
Use latest aruba

deivid-rodriguez:
I'd like to use aruba's master for our specs because I added [a change](https://github.com/cucumber/aruba/pull/675) that I'm finding useful when working with simplecov's cucumber scenarios.

I also removed an extension recently added to aruba, because otherwise specs won't run against aruba 1.0. I thought Aruba 1.0 was fixing the issues, but then I noticed that even without upgrading aruba, the scenarios will pass anwyays without the extension. So, I'm not sure why the aruba extension is needed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ gem "simplecov-html", github: "colszowka/simplecov-html"
 
 group :development do
   gem "aruba", "~> 0.14"
-  gem "capybara", :github => "teamcapybara/capybara"
+  gem "capybara", github: "teamcapybara/capybara"
   gem "cucumber", "~> 3.1"
   gem "cuprite", "0.8"
   gem "rake", "~> 13.0"
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ source "https://rubygems.org"
 gem "simplecov-html", github: "colszowka/simplecov-html"
 
 group :development do
-  gem "aruba", "~> 0.14"
+  gem "aruba", github: "cucumber/aruba"
   gem "capybara", github: "teamcapybara/capybara"
   gem "cucumber", "~> 3.1"
   gem "cuprite", "0.8"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,6 +5,18 @@ GIT
     simplecov-html (0.11.0.beta2)
 
 GIT
+  remote: https://github.com/cucumber/aruba.git
+  revision: 8b7dc956daf47051237700488e8d866fb229db29
+  specs:
+    aruba (1.0.0.pre.alpha.5)
+      childprocess (~> 3.0)
+      contracts (~> 0.16.0)
+      cucumber (>= 2.4, < 4.0)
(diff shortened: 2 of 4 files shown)


PragTob:
@deivid-rodriguez thanks!

Let me double check on the extension... I added it.. 3 days ago? See: https://github.com/cucumber/aruba/issues/699

... Ah I believe it might be working because the scenarios in question are tagged `@wip` right now because they are flaky due to #815 but after that's fixed I really want them in the general rotation (+ I'd like to test against different rails version) so removing it right now I wouldn't be too much of a fan

deivid-rodriguez:
@PragTob Thanks for the context!

So I reviewed the monkeypatch and I believe it was correct but being applied to the wrong method. I fixed it and now I think the scenarios should pass consistently. In addition to that, what I did was to remove the hard failure on aruba 1.0, and replace it with a TODO note.

Let me know what you think.

PragTob:
@deivid-rodriguez thanks I'll check it out! I don't believe the scenarios will pass consistently though, as imo it's a bug in our merging that needs fixing :)

deivid-rodriguez:
Oh, I see! Then I'll restore the `@wip` tags and the mention to the ticket until we can fix that.

PragTob:
@deivid-rodriguez sounds great :ok_hand: 

I just double checked and yeah still happening (took me a lot of retries though):

```
tobi@speedy:~/github/simplecov(tests/latest_aruba)$ while cucumber features/parallel_tests.feature:8; do :; done
Using the default profile...
...F

(::) failed steps (::)


expected: 9
     got: 8

(compared using ==)
 (RSpec::Expectations::ExpectationNotMetError)
./features/step_definitions/html_steps.rb:14:in `block (3 levels) in <top (required)>'
./features/step_definitions/web_steps.rb:5:in `block in with_scope'
./features/step_definitions/web_steps.rb:5:in `with_scope'
./features/step_definitions/html_steps.rb:12:in `block (2 levels) in <top (required)>'
./features/step_definitions/html_steps.rb:11:in `each'
./features/step_definitions/html_steps.rb:11:in `/^I should see the groups:$/'
./features/step_definitions/parallel_tests_steps.rb:4:in `"I should see the line coverage results for the parallel tests project"'
features/parallel_tests.feature:16:in `Then I should see the line coverage results for the parallel tests project'

Failing Scenarios:
cucumber features/parallel_tests.feature:8 # Scenario: Running it through parallel_tests produces the same results as a normal rspec run

```

deivid-rodriguez:
Right.

Regarding the change in the monkeypatch, I investigated more and it's not that the previous monkeypatch was incorrect. It was correct on aruba 0.14, but on aruba 1.0, the "I successfully run" step [no longer calls `run_command`](https://github.com/cucumber/aruba/commit/baa6cfacb2c878512f5452c9a0346dff64839610) so the monkeypatch no longer applies. Monkeypatching `run_command_and_stop` instead does the trick.

This should be ready now.

PR #826, opened 2020-01-18
Stop generally symbolizing all the keys

PragTob:
A lot of them we convered back to strings anyhow, plus with
frozen string literal the gains are negligible anyhow.

Also, always symbolizing everything in the JSON.parse was a bit
of an overkill especially when you imagine a code base with lots
of files and lots of branches we could easily contribute to
blowing up memory considerably/beyond an accepatable threshhold.

### DIFF
--- a/lib/simplecov/source_file.rb
+++ b/lib/simplecov/source_file.rb
@@ -12,7 +12,7 @@ module SimpleCov
     attr_reader :coverage
 
     def initialize(filename, coverage)
-      @filename = filename.to_s
+      @filename = filename
       @coverage = coverage
     end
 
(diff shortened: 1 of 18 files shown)


lib/simplecov/source_file.rb
```diff
@@ -199,9 +199,8 @@ def restore_ruby_data_structure(structure)
       # put them through here.
       return structure if structure.is_a?(Array)
 
-      # as of right now the keys are still symbolized
```
deivid-rodriguez: The docs on top of this method also talk about key symbolization so should probably be updated too?

PragTob: Great catch!

PragTob:
I've been going mad as to why these 2 specs fail but they don't fail on master... it's because I fixed the rake task and they just weren't executed before :man_facepalming: 

(tis what happens when you go have lunch inbetween coding)

PragTob:
Yes I should leave this open longer but I need this for... reasons ;)

PR #827, opened 2020-01-18
Disable Bundler the Aruba way

mvz:
I removed the `@wip` tags just to demonstrate that this works with Aruba master. That bit 
should not be merged (hence the draft status).

This method of disabling Bundler is indeed not working with Aruba 0.14.14.


deivid-rodriguez:
I think it does make sense to remove the `@wip` tags here since this issue is the reason why they are there I believe.

deivid-rodriguez:
Oh, no, I see it's because of a race condition, sorry about that. :pray:

PragTob:
@mvz thanks a lot! :tada: 

Yup that's the race condition they're there for which I might or not might try to dig into at some point in time but it seems daunting :scream: 

So if we could reinstate those, that'd be great. The failing build war JRuby version that is somehow flaky on travis (sometimes doesn't know how to install it...) - restarted that build!

mvz:
Ok, seeing that the specs pass I'll reinstate the `@wip` tags.

PragTob:
Thanks a lot @mvz ! :green_heart: 

IMG_20190130_181749

PR #831, opened 2020-01-18
Migrate CI to github actions

deivid-rodriguez:
This should fix the unstability we're having in TravisCI with JRuby at the moment.


deivid-rodriguez:
@PragTob Have you seen the current failure before? I think it's a flaky failure not related to this PR?

deivid-rodriguez:
The flaky is this one: https://github.com/colszowka/simplecov/runs/396912955

PragTob:
@deivid-rodriguez I've never seen that failure before :( It might be though/timing related as I introduced this step and it's only used (so far) in the branch coverage feature I built :scream: 

It might very well be flaky/timing related:

```
When /^I open the detailed view for "(.+)"$/ do |file_path|
  click_on(file_path)

  header_text = page.find(".header h3", visible: true).text
  expect(header_text).to eq file_path
end
```

Aka click hasn't yet gone through and we're already checking against the text because the check isn't patient. So likely not a problem of the CI but of me not writing it patient :sweat_smile: 

I'll fix it!

deivid-rodriguez:
Right, I think that's a classic capybara flaky, and I _think_ that you can fix it by using the `have_selector` matcher, that automatically waits for the content to appear.

PragTob:
@deivid-rodriguez yup exactly, fix is up on #834 - I'll wait for it to pass CI there, merge it and then rebase this one against master

.travis.yml
```diff
@@ -1,25 +0,0 @@
...
-    - rvm: jruby-9.2.9.0
-      env: JRUBY_OPTS=--debug
-      jdk: openjdk8
-    - rvm: ruby-head
-    - rvm: jruby-head
-      env: JRUBY_OPTS=--debug
-      jdk: openjdk8
-  allow_failures:
-    - rvm: ruby-head
-    - rvm: jruby-head
```
PragTob: I haven't looked into github actions so far - is there a way for us to get back our checks against head? Truthfully because they were partially broken we didn't really use them so they're not a deal breaker but would be good to know

.travis.yml
```diff
@@ -1,25 +0,0 @@
...
-    - rvm: 2.5.7
-    - rvm: 2.6.5
-    - rvm: 2.7.0
-    - rvm: jruby-9.2.9.0
-      env: JRUBY_OPTS=--debug
-      jdk: openjdk8
-    - rvm: ruby-head
-    - rvm: jruby-head
-      env: JRUBY_OPTS=--debug
-      jdk: openjdk8
```
PragTob: Similar question, can one have jdk switching on github actions/do you know? We had some trouble where we had to pin it to older JDKs due to problems with new JDKs and I'd dislike not having control over that - or well at least I'd be concerned.

.travis.yml
deivid-rodriguez: Ouch, I forgot that! There's no currently "built-in way" to do that. So the options are:

* Keep TravisCI for ruby-head for the moment.
* Add a separate workflow that installs `rvm`, and ruby-head on top of it.
* Use the official nightly docker image: https://github.com/ruby/ruby-docker-images.

.travis.yml
deivid-rodriguez: That's a good question I don't know the answer too, but I'll research.

.travis.yml
PragTob: Cool thanks, as I said not a deal breaker for me so I'm happy to go ahead without any of this but it's great to know that we got options :ok_hand:

PragTob:
rebased and force pushed with flaky fix :angel:

.travis.yml
deivid-rodriguez: I migrated the old `ruby-head` and `jruby-head` old TravisCI entries to keep using RVM under github actions.

.travis.yml
deivid-rodriguez: I fixed this by manually installing the requested jdk.

deivid-rodriguez:
@PragTob I addressed the issues. New matrix under Github Actions should be fully equivalent to the previous one on TravisCI.

.github/workflows/stable.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Install bundler
+        run: gem i bundler
+
+      - name: Install dependencies
+        run:  bundle install --without benchmark --jobs=3 --retry=3
+
+      - name: Run tests (MRI)
+        run: bundle exec rake
+        if: matrix.ruby == 'ruby-head'
```
PragTob: I'm confused, this looks like it would never run our tests since we have no `ruby-head` entry in the ruby matrix here?

.github/workflows/stable.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Run tests (JRuby)
+        run: JRUBY_OPTS=--debug bundle exec rake
+        if: matrix.ruby == 'jruby-head'
```
PragTob: same here we don't seem to have jruby-head in here just in the other file (and it should run for normal JRuby) or am I missing something?

PragTob:
@deivid-rodriguez either I don't understand how this works or it's bugged :) See questions above :grin:

.github/workflows/stable.yml
deivid-rodriguez: Oops, nice catch :sweat_smile:.

deivid-rodriguez: I pushed a fix.

deivid-rodriguez:
It was just some copy-paste errors, it should make some sense now.

PragTob:
Thank you!

WhatsApp Image 2019-11-11 at 08 03 15

PR #835, opened 2020-01-20
Adapt old resultset.jsons instead of just warning

PragTob:
After discussion with @deivid-rodriguez I realized that adapting
the result isn't actually that difficult and it's actually
a better solution to my current problem (that the warning
gets printed multiple times).

Yes, one could have moved the code around so that wrong format
results wouldn't even be written back, but Deivid was right
it might be useful for some people to keep the result and it's
actually not more complicated than sorting them out.

So this change fixes the multiple warnings as well as keeping
old results around.

Also moved the checking/changing to a fairly central place to
safeguard against old results flying in from somewhere else.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+0.18.0.beta3 (unreleased)
+========================
+
+## Enhancements
+* Instead of ignoring old `.resultset.json`s that are inside the merge timeout, adapt and respect them
+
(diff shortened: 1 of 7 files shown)


PragTob:
@deivid-rodriguez as before a look would be appreciated if you got the time :) If not, no biggie of course. Would want to release the fix this evening (CET) before it annoys more people :angel:

PR #836, opened 2020-01-23
Fix Github Actions badge link

deivid-rodriguez:
Since I'm labeling the badge as "stable", I wanted to actually link to the "stable" workflow, which should always be green, while the other might not.


deivid-rodriguez:
It sounds like we exchanged one flaky for another :(

PragTob:
@deivid-rodriguez thanks!

Not sure what you mean with flaky, haven't seen one as you presumably reran it. If you see something flaky please post an issue :)

deivid-rodriguez:
I rerun it, and the flaky is now lost, at least I can no longer find it. I'll post an issue if it reappers.

PR #848, opened 2020-01-30
tobi blindly tries to docker/github workflow

PragTob:


PragTob:
I was hoping I could blindly insert docker images in the runs-on, turns out I can not and it's more involved so I won't tackle this right now :(

deivid-rodriguez:
It's like this: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idcontainer

deivid-rodriguez:
@PragTob If we're good with removing the `jdk8` requirement, and instead use whatever java version comes with the `jruby:9.2` image, like https://github.com/colszowka/simplecov/pull/848/commits/0318a6269db8e45f25708c0723e565529c3b2118 suggests, it seems simpler to remove the `apt-get install` step from the current workflow and instead use whatever `java` version comes with `ubuntu`. Right?

PragTob:
@deivid-rodriguez is there a default JRE installed there?

I'm not sure if we're fine without it yet, I know there were/are problems but not sure if they are fixed. I fear they might not be, but it's worth trying.

At the time I just wanted to know/make sure that we could get an older JDK if we needed to because I've been through the pain before.

deivid-rodriguez:
Yeah, I'm pretty sure it works because the first version of the workflow didn't include it. I added it because you pointed out the difference from the previous TravisCI setup in https://github.com/colszowka/simplecov/pull/831#discussion_r368437502.

PragTob:
yeah I remember and thank you :) I just wanted to make sure we could switch if needbe as historicaly there were problems relating to the JVM version but I think if they pop up again I/we know how to tackle them so I'd be fine momentarily running without them.

PragTob:
yay no chrome executable... well let's hope for the other branch but yeah that was to be expected :sweat_smile:

PragTob:
Closed in favor of #850

PR #850, opened 2020-01-31
Try without installing jdk8 manually

PragTob:
At least that's what I think deivid meant :)

https://github.com/colszowka/simplecov/pull/848#issuecomment-580640435


deivid-rodriguez:
Yes, that's exactly what I meant :)

PR #852, opened 2020-02-01
Refactor CoverageStatistics class & specs and fix typos

ddzz:
Simple refactor of `SimpleCov::CoverageStatistics` that makes the code slightly more efficient and readable. Also fixes a spec (along with some clean up).

I also searched for typos using [cspell](https://www.npmjs.com/package/cspell). I manually fixed the obvious typos that were reported (and ignored tons of false positive matches).

### DIFF
--- a/lib/simplecov/coverage_statistics.rb
+++ b/lib/simplecov/coverage_statistics.rb
@@ -19,7 +19,7 @@ module SimpleCov
           [
             covered + file_coverage_statistics.covered,
             missed + file_coverage_statistics.missed,
-            # gotta remultiply with loc because files have different strenght and loc
+            # gotta remultiply with loc because files have different strength and loc
             # giving them a different "weight" in total
             total_strength + (file_coverage_statistics.strength * file_coverage_statistics.total)
           ]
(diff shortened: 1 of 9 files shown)


lib/simplecov/coverage_statistics.rb
```diff
@@ -35,22 +35,22 @@ def initialize(covered:, missed:, total_strength: 0.0)
...
+      return 100.0 if @missed.zero?
 
-      covered * 100.0 / total
+      covered * 100.0 / @total
     end
 
-    def compute_strength(total_strength, total)
-      return 0.0 if total.zero?
+    def compute_strength(total_strength)
+      return 0.0 if @total.zero?
```
PragTob: The usage of parameters here (and rest of the file) was intentional. That way it's clearer to me what part of the computation relies on which values just by looking at the method signature/seeing their calls in initialize :)

I know one could debate it isn't very OO and if they were public it wouldn't be like this, but personally I like them better like this :)

spec/coverage_statistics_spec.rb
```diff
@@ -3,7 +3,7 @@
 require "helper"
 
 RSpec.describe SimpleCov::CoverageStatistics do
-  describe ".new" do
+  describe "::new" do
```
PragTob: ? That's not the convention I usually follow in Ruby. For me it's `.class_method` and `#instance_method` :)

spec/coverage_statistics_spec.rb
```diff
@@ -16,20 +16,20 @@
     end
 
     it "can omit the total strength defaulting to 0.0" do
-      statistics = described_class.new(covered: 4, missed: 6, total_strength: 0.0)
+      statistics = described_class.new(covered: 4, missed: 6)
```
PragTob: :+1:

spec/coverage_statistics_spec.rb
```diff
@@ -16,20 +16,20 @@
...
 
       expect(statistics.strength).to eq 0.0
     end
 
     it "can deal with it if everything is 0" do
-      statistics = described_class.new(covered: 0, missed: 0, total_strength: 0.0)
+      statistics = empty_statistics
```
PragTob: technically yes, while the test description says "everything is 0" I like to see that everything is 0 that makes it immedeatly a bit clearer (for this one explicit test) and it's not such a huge amount of duplication imo.

spec/coverage_statistics_spec.rb
ddzz: The `::` convention is used [in Ruby Docs](https://ruby-doc.org/core-2.2.0/BasicObject.html) for public class methods (see the list of methods on the left), but I do think the RSpec convention is to use `.`. I'll change it back. Thanks for reviewing!

PR #855, opened 2020-02-06
Rebase: Fix at_exit hook to cooperate with Minitest

PragTob:
This is me trying to resurrect #756

### DIFF
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -189,6 +189,14 @@ module SimpleCov
       end
     end
 
+    def at_exit_behvior
+      # If we are in a different process than called start, don't interfere.
+      return if SimpleCov.pid != Process.pid
+
+      # If SimpleCov is no longer running then don't run exit tasks
+      SimpleCov.run_exit_tasks! if SimpleCov.running
+    end
+
     # @api private
     #
     # Called from at_exit block
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+unreleased
+===========
+
+## Bugfixes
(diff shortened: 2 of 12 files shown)


PragTob:
@adam12 if you get to try this out to see if I didn't accidentally break the fix I'd be very happy :) But the feature works so I hope for the best :crossed_fingers:

adam12:
Looks great. I'm going to give it a try sometime this week.

I never could figure out why the coverage changed on #756 between Ruby versions.

PragTob:
@adam12 ah I squashed all the commits, basically in the `test_helper.rb` of the `begin .. end` the begin was marked as relevant in 2.5 but not the others which I consider a coverage bug. I solved it by ignoring `test_helper.rb` (not interested in its coverage anyhow).

Let me know what you find once you try it out :crossed_fingers: And thanks for your help with this :green_heart:

PragTob:
@adam12 :wave: Do you know if you'll get to trying this out? No pressure, would just be nice since you had the initial problems. I'm looking to merge it some time this week.

adam12:
> Do you know if you'll get to trying this out?

_Looks_ to be functioning correctly here. At least to say, I get coverage report at the end of the test run and not immediately before the tests start running as was happening previously.

PragTob:
@adam12 thanks for trying it out! Will merge :grin:

lib/simplecov.rb
```diff
@@ -189,6 +189,14 @@ def exit_status_from_exception
       end
     end
 
+    def at_exit_behvior
```
Yoshiji: typo at `behvior`? `behavior`?

PragTob: good catch, thank you!

PR #856, opened 2020-02-06
Amend 0.18 changelog with notice about code climate test reporter

PragTob:
I didn't add this information as to my mind we didn't change any API documented anywhere and hence it shouldn't appear in a Changelog.

However, this affects people and they have to find out for themselves so notifying them of this might be a good thing. I still couldn't bring myself to put it under breaking changes as technically we haven't broken anything that we ever communicated as working.

So a link to https://github.com/codeclimate/test-reporter/issues/413 seems to be helpful.


PragTob:
I'm still not too sure about this, @deivid-rodriguez & others would be happy about your input as to where this should end up and how it should be worded.

deivid-rodriguez:
In my opinion, this is definitely worth mentioning, yeah, no matter whether we consider this a breaking change or not.

PR #858, opened 2020-02-12
Be more explicit about choosing test projects

PragTob:
With the incoming introduction of more test projects, I believe
it's better to explicitly step into them rather than relying
on the cucumber annotations to get into one project or the
other.

Yes it's more code, but less magic and more explicitness.

Drive by renamed the parallel tests project, no need to extra
call out that it's a project.


deivid-rodriguez:
Totally agreed, looks much better :+1:.

PR #863, opened 2020-02-13
Allow simplecov-html 0.12

jaredbeck:
The [changelog for 0.12](https://github.com/colszowka/simplecov-html/blob/master/CHANGELOG.md#0120-2020-02-12) does not mention any breaking changes.

### DIFF
--- a/simplecov.gemspec
+++ b/simplecov.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |gem|
   gem.required_ruby_version = ">= 2.4.0"
 
   gem.add_dependency "docile", "~> 1.1"
-  gem.add_dependency "simplecov-html", "~> 0.11"
+  gem.add_dependency "simplecov-html", [">= 0.11", "< 0.13"]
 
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
It's already allowed:

```
> Gem::Requirement.new("~> 0.11").satisfied_by?(Gem::Version.new("0.12.0"))
=> true
```

My guess is that you were using 0.18.1 where this had not yet been releaxed.

jaredbeck:
> My guess is that you were using 0.18.1 where this had not yet been releaxed.

Thanks, David. That's exactly what happened.

jaredbeck:
Unrelated, I am surprised that rubygems considers 0.12 to be compatible (per `~>`) with 0.11, because SemVer allows breaking changes between 0.x versions.

deivid-rodriguez:
`~> x.y` is the best approximation to semver rubygems has, but it's not a "real" semver requirement as you point out. `~> x.y` is just a shortcut for `>= x.y, < (x+1).0`

Issue #853, opened 2020-02-04
Merged SimpleCov Reporting Differences Between v0.17.1 and v0.18.0


Justin-W:
There has been a major drop in the coverage reported by simplecov between v0.17.1 and v0.18.0. And v0.18.1 reports the same as v0.18.0.

From what I can tell, I suspect that the issue relates to some difference in the report-merging functionality, since my coverage data involves extensive report merging, and the previous and new coverage reports seem to differ only in the number of hits for each line, but not in which lines have a hit count.

## Test suite configuration

I am using rspec (via rake tasks) to run my test suite, and `simplecov-cobertura` to format the results as xml (although the html report has also been affected, but is now showing 0% coverage instead of the reduced, but non-zero, coverage that the xml report shows; will file a second bug about that).

Included within my rspec test suite are some specs that call my gem from the CLI, with simplecov configured to generate separate report files, with filenames of the format 'bin_terra_<spec_index>_<process_pid>'. The separate calls to the gem represent different subcommands and/or different CLI flags, and are expected to generate dozens of hits to non-conditional lines, and fewer hits on conditionally-called lines.

Since the test suite often takes between 20-45 minutes to run, I have set `SimpleCov.merge_timeout(3600)` in my `.simplecov` file.

Nothing has changed except for the simplecov and simplecov-html versions (i.e. no changes to the covered code, nor the rspec code, nor the CLI inputs, nor the number of simplecov reports being merged).

The test reports that I am reviewing are all generated within clean CI environments.

## Expected results

Using v0.17.1, I was seeing 82% coverage of my 'lib' package.

## Actual results

Using v0.18.0 (same results with v0.18.1), I am seeing 15% coverage of my 'lib' package, and many/most lines are reporting fewer line hits than previously (see before and after report excerpts below).

Interesting observations from the excerpts below:

- Note the line hit counts of line 8 in the excerpts below (the `def initialize` method declaration): 25 vs 1.
- Note the line hit counts of line 9 in the excerpts below (the first line of the `initialize` method): 39 vs 13.
- Note the line hit counts of line 77 in the excerpts below (the `def deploy_apacheds` method declaration): 25 vs 1.
- Note the line hit counts of line 80 in the excerpts below (the first line of the `deploy_apacheds` method): 16 vs 0.
- Given that the # of tests executed, the test code, and the covered code are all unchanged, there should be no differences in the hit counts of lines 8, 9, 77, or 80 between the 2 reports below.
[...]

PR #864, opened 2020-02-13
Cli acceptance testing test stub enhancements and #853 reproducability

Justin-W:
The PR extends simplecov PR#859 with additional test code and specs, and adds (currently-disabled) code to reproduce simplecov issue #853, as requested by @PragTob.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,10 +18,13 @@ tmtags
 
 ## PROJECT::GENERAL
 /coverage
+coverage/
+coverage.*/
 rdoc
 pkg
 tmp
 .rvmrc
+/.idea/
 
 ## PROJECT::SPECIFIC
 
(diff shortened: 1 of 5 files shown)


Justin-W:
@PragTob I fixed several issues which caused CI failures, but I don't know how to fix this most recent failure. It seems unrelated to my changes. 😕

PragTob:
@Justin-W yeah seems like some kind of dependency thing that I should fix upstream I guess. Don't worry about it.

So, does the error only occur when using the cobertura formatter? That'd be mightily weird and outside of the scope of simplecov itself but also interesting likely... question is if it'd also occur with other formatters.

I think right now it also wouldn't work as we'd need to add the gem here or at least in one of the projects.

PR #865, opened 2020-02-14
Convert source file encoding to UTF-8

Tietew:
Fixes colszowka/simplecov-html#91

## Summary
- Open the source file in encoding UTF-8 instead of ASCII-8BIT.
- Check for magic comment and call `IO#set_encoding` to convert encoding to UTF-8.

## Background
`File.open(..., "rb")` will set its encoding to ASCII-8BIT instead of script encoding (UTF-8).
This means all lines are ASCII-8BIT.

In simplecov-html, following line
https://github.com/colszowka/simplecov-html/blob/7540373ed44ccd43d1347b775a69673297cc8f90/views/source_file.erb#L48
will try to convert `src` from ASCII-8BIT to UTF-8!
Therefore, all non-ASCII characters will be replaced by Unicode Replacement Character.

### DIFF
--- a/lib/simplecov/source_file.rb
+++ b/lib/simplecov/source_file.rb
@@ -25,7 +25,7 @@ module SimpleCov
     def src
       # We intentionally read source code lazily to
       # suppress reading unused source code.
-      @src ||= File.open(filename, "rb", &:readlines)
+      @src ||= load_source
     end
     alias source src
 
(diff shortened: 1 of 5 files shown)


PragTob:
:wave: 

Hi there, thanks for your fix! :green_heart: 

I'll have to mull the best approach over in my head a bit. I think this is pretty great and sophisticated but I think I need some time to think about and play with it.

Thanks!

lib/simplecov/source_file.rb
```diff
@@ -175,6 +175,32 @@ def build_no_cov_chunks
       end
     end
 
+    def load_source
+      lines = []
+      # The default encoding is UTF-8
+      File.open(filename, "rb:UTF-8") do |file|
+        line = file.gets
+
+        # Check for shbang
```
tycooon: Should be `shebang`.

PragTob:
Thanks a lot, I think this is the right approach! I created #866 to build upon this and maybe still add some more specs (as files can have many forms I'm worried about something crashing while we try to read it :sweat_smile: ).

Goal is to write the cuke this evening and package up a release for early next week or maybe even this evening depending on how long I stay out today ;)

PragTob:
Closing in favor of #866

lib/simplecov/source_file.rb
Tietew: Thanks for correction.

PR #866, opened 2020-02-16
Convert utf8

PragTob:
Based on #865 

I had a bit of trouble grasping the solution, so needed to play with myself and add some more tests to make sure nothing does/did explode.

Still probably want to add one acceptance test to make sure the html formatter doesn't blow it up for some reason.


PragTob:
Well no release today due to me wanting to report https://github.com/jruby/jruby/issues/6093 - see you back here tomorrow ;)

PR #874, opened 2020-02-24
Accessor and guard for external at exit handler

adam12:
Fixes #873

PR #875, opened 2020-02-24
Alternative at exit guard

PragTob:
#874 + acceptance test


PragTob:
Rails 6 of course doesn't work with Ruby 2.4. Fixing...

Issue #414, opened 2015-08-25
Picking up coverage during fork


ioquatix:
The following coverage: https://coveralls.io/builds/3383271 is not picking up coverage of code during fork.

https://coveralls.io/builds/3383271/source?filename=lib%2Fprocess%2Fdaemon%2Fcontroller.rb

Not sure exactly how to debug this or help figure out what is going on except to ask here.

bf4:
Hi @ioquatix ! Did you set `SimpleCov.pid = $$` in your `.simplecov`? or elsewhere? Have you tried setting a command name?  Are you running parallel tests, or there's just a fork somewhere? 

SimpleCov version etc, please :) https://github.com/colszowka/simplecov/blob/master/CONTRIBUTING.md#reporting-issues

ioquatix:
The .simplecov file: https://github.com/ioquatix/process-daemon/blob/master/.simplecov

I don't believe I'm doing anything like `SimpleCov.pid = $$` anywhere.

It uses whatever is the latest version of SimpleCov, e.g. https://github.com/ioquatix/process-daemon/blob/master/Gemfile - this is running on Travis.

Yes, it's double forking to start a daemon process.

bf4:
Nah, you're doing it right with the pid.

bf4:
other possibility is changing the command name in the fork, just as you would with parallel tests...

ioquatix:
Okay so like `Process.setproctitle` ?

bf4:
oh, I meant `fork { SimpleCov.command_name = "i'm a fork" }`

ioquatix:
I tried command_name, ti didn't help. I also tried changing my fork code like so

```
			def spawn
				@daemon.prefork
				@daemon.mark_log

				fork do
					Process.setsid
					exit if fork

					SimpleCov.start

					# ... snip ...
```

It didn't help. It's like after it enters the fork block coverage is not recorded. What can I do to check?

bf4:
Puts Coverage.result, pid

Or 

B mobile phone

PragTob:
I'm not sure what to do here and if forking can be made to work with simplecov/coverage. At the elast we should probably document it in the README.

PR #881, opened 2020-03-03
Add subprocess handling to simplecov

robotdana:
Using Process.fork (whether using the Parallel gem or directly)
creates code that was invisible to SimpleCov. by starting SimpleCov
within the subprocess with its own command name and etc we can
see that code :)

This also adds documentation for what to do when using Process.spawn or
similar.

fixes: #414 and probably others

---

This seems the more correct place to fix this than my previous attempt here: https://github.com/grosser/parallel/issues/275

Testing was a bit of a mess, perhaps you have some ideas, 
Also what are your thoughts on this being enabled by default or not?

Thanks :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -623,13 +623,13 @@ to the `SimpleCov.command_name`, with results that can be merged together using 
 To configure this, use `.at_fork`.
 
 ```ruby
-SimpleCov.enable_for_subprocesses = true
+SimpleCov.enable_for_subprocesses true
 SimpleCov.at_fork do |pid|
   # This needs a unique name so it won't be ovewritten
   SimpleCov.command_name "#{SimpleCov.command_name} (subprocess: #{pid})"
   # be quiet, the parent process will be in charge of output and checking coverage totals
   SimpleCov.print_error_status = false
-  SimpleCov.formatter = SimpleCov::Formatter::SimpleFormatter
+  SimpleCov.formatter SimpleCov::Formatter::SimpleFormatter
   SimpleCov.minimum_coverage 0
   # start
   SimpleCov.start
--- a/README.md
+++ b/README.md
@@ -614,6 +614,50 @@ namespace :coverage do
 end
 ```
 
+## Running simplecov against subprocesses
+
+SimpleCov will cover code run using `Process.fork` automatically, appending `" (subprocess #{pid})"`
+to the `SimpleCov.command_name`, with results that can be merged together using SimpleCov's merging feature.
+
+To configure this, use `.at_fork`.
+
+```ruby
+SimpleCov.at_fork do |pid|
+  # This needs a unique name so it won't be ovewritten
+  SimpleCov.command_name "#{SimpleCov.command_name} (subprocess: #{pid})"
+  # be quiet, the parent process will be in charge of using the regular formatter and checking coverage totals
+  SimpleCov.print_error_status = false
+  SimpleCov.formatter = SimpleCov::Formatter::SimpleFormatter
+  SimpleCov.minimum_coverage 0
+  # start
+  SimpleCov.start
+end
+```
+
+NOTE: SimpleCov must have already been started before `Process.fork` was called.
+
+### Running simplecov against spawned subprocesses
+
+Perhaps you're testing a ruby script with `PTY.spawn` or `Open3.popen`, or `Process.spawn` or etc.
+SimpleCov can cover this too.
+
+Add a .simplecov_spawn file to your project root
+```ruby
+# .simplecov_spawn
+require 'simplecov' # this will also pick up whatever config is in .simplecov
+                    # so ensure it just contains configuration, and doesn't call SimpleCov.start.
+SimpleCov.at_fork.call(Process.pid)
+SimpleCov.start
+```
+Then, instead of calling your script directly, like:
+```ruby
+PTY.spawn('my_script.rb') do # ...
+```
+Use bin/ruby to require the new .simplecov_spawn file, then your script
+```ruby
+PTY.spawn('ruby -r.simplecov_spawn my_script.rb') do # ...
+```
+
 ## Running coverage only on demand
 
 The Ruby STDLIB Coverage library that SimpleCov builds upon is *very* fast (on a ~10 min Rails test suite, the speed
--- a/README.md
+++ b/README.md
@@ -616,16 +616,18 @@ end
 
 ## Running simplecov against subprocesses
 
-SimpleCov will cover code run using `Process.fork` automatically, appending `" (subprocess #{pid})"`
+`SimpleCov.enable_for_subprocesses` will allow SimpleCov to observe subprocesses starting using `Process.fork`.
+This modifies ruby's core Process.fork method so that SimpleCov can see into it, appending `" (subprocess #{pid})"`
 to the `SimpleCov.command_name`, with results that can be merged together using SimpleCov's merging feature.
 
 To configure this, use `.at_fork`.
 
 ```ruby
+SimpleCov.enable_for_subprocesses = true
 SimpleCov.at_fork do |pid|
   # This needs a unique name so it won't be ovewritten
   SimpleCov.command_name "#{SimpleCov.command_name} (subprocess: #{pid})"
-  # be quiet, the parent process will be in charge of using the regular formatter and checking coverage totals
+  # be quiet, the parent process will be in charge of output and checking coverage totals
   SimpleCov.print_error_status = false
   SimpleCov.formatter = SimpleCov::Formatter::SimpleFormatter
   SimpleCov.minimum_coverage 0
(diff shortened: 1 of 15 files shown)


robotdana:
(I'm not clear why test 2.4.9 didn't work, it doesn't seem related?)

README.md
```diff
@@ -614,6 +614,50 @@ namespace :coverage do
...
+### Running simplecov against spawned subprocesses
+
+Perhaps you're testing a ruby script with `PTY.spawn` or `Open3.popen`, or `Process.spawn` or etc.
+SimpleCov can cover this too.
+
+Add a .simplecov_spawn file to your project root
+```ruby
+# .simplecov_spawn
+require 'simplecov' # this will also pick up whatever config is in .simplecov
+                    # so ensure it just contains configuration, and doesn't call SimpleCov.start.
```
PragTob: :+1: great docs

lib/simplecov.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "English"
+require_relative "simplecov/process"
```
PragTob: Not sure requiring this by default is our best option. While I'm all for providing a good out of the box experience, I also don't want to freedom patch core - I see that it seems necessary for what you want to achieve unless we'd put more burden on the user. What do you think about telling users themselves to require this? (+ I'd probably name it `process_patch` or something to indicate what we're doing)

lib/simplecov/configuration.rb
```diff
@@ -196,6 +196,40 @@ def at_exit(&block)
...
+    #     end
+    #
+    def at_fork(&block)
+      @at_fork = block if block_given?
+      @at_fork ||= lambda { |pid|
+        # This needs a unique name so it won't be ovewritten
+        SimpleCov.command_name "#{SimpleCov.command_name} (subprocess: #{pid})"
+        # be quiet, the parent process will be in charge of using the regular formatter and checking coverage totals
+        SimpleCov.print_error_status = false
+        SimpleCov.formatter = SimpleCov::Formatter::SimpleFormatter
```
PragTob: do we need to set the formatter here?

lib/simplecov/configuration.rb
```diff
@@ -196,6 +196,40 @@ def at_exit(&block)
...
+        SimpleCov.minimum_coverage 0
```
PragTob: Do we need to set this? It should only be evaluated in the end... which now occurs to me we might not be detecting correctly because SimpleCov isn't used to this yet but rather checks for parallel processes.

spec/simplecov_spec.rb
```diff
@@ -270,6 +270,31 @@
...
+      end
+
+      it "starts coverage of subprocesses" do
+        SimpleCov.command_name "RSpec"
+        SimpleCov.start
+
+        a_void_method_call_that_wont_warn # this process
+
+        pid = Process.fork do
+          a_void_method_call_that_wont_warn # that process
```
PragTob: It's unclear to me why we need those method calls - to generate data so that we have a result in the end?

spec/simplecov_spec.rb
```diff
@@ -270,6 +270,31 @@
...
+        clear_mergeable_reports
+      end
+
+      def a_void_method_call_that_wont_warn
+        true
+      end
+
+      it "starts coverage of subprocesses" do
+        SimpleCov.command_name "RSpec"
+        SimpleCov.start
```
PragTob: This affects global state (sadly in the design of SimpleCov) - which I guess would make writing another test for this hard and might lead to funky interactions with other tests down the road. But yes, I know you said testing was tricky I'm trying to come up with something better :thinking:

PragTob:
Ah forgot, I'm also not sure why the test run failed. I'll save it's error message here and rerun it.

```
336 examples, 0 failures

Randomized with seed 39135

	from /home/runner/work/simplecov/simplecov/lib/simplecov.rb:388:in `process_coverage_result'
	from /home/runner/work/simplecov/simplecov/lib/simplecov.rb:108:in `result'
	from /home/runner/work/simplecov/simplecov/lib/simplecov/configuration.rb:196:in `block in at_exit'
	from /home/runner/work/simplecov/simplecov/lib/simplecov.rb:215:in `run_exit_tasks!'
	from /home/runner/work/simplecov/simplecov/lib/simplecov.rb:203:in `at_exit_behavior'
	from /home/runner/work/simplecov/simplecov/lib/simplecov/defaults.rb:27:in `block in <top (required)>'
/home/runner/.rubies/ruby-2.4.9/bin/ruby -I/home/runner/.rubies/ruby-2.4.9/lib/ruby/gems/2.4.0/gems/rspec-core-3.9.1/lib:/home/runner/.rubies/ruby-2.4.9/lib/ruby/gems/2.4.0/gems/rspec-support-3.9.2/lib /home/runner/.rubies/ruby-2.4.9/lib/ruby/gems/2.4.0/gems/rspec-core-3.9.1/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb failed
##[error]Process completed with exit code 1.
```

PragTob:
Failure definitely seems related, now it's happening in 2.6.5 - I guess it's some kind of race condition introduced by the test calling `.result` - moving it to cukes would also get rid off that :)

```
 /home/runner/.rubies/ruby-2.6.5/bin/ruby -I/home/runner/.rubies/ruby-2.6.5/lib/ruby/gems/2.6.0/gems/rspec-core-3.9.1/lib:/home/runner/.rubies/ruby-2.6.5/lib/ruby/gems/2.6.0/gems/rspec-support-3.9.2/lib /home/runner/.rubies/ruby-2.6.5/lib/ruby/gems/2.6.0/gems/rspec-core-3.9.1/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb

Randomized with seed 17877
/home/runner/work/simplecov/simplecov/lib/simplecov.rb:399:in `result': coverage measurement is not enabled (RuntimeError)
	from /home/runner/work/simplecov/simplecov/lib/simplecov.rb:399:in `adapt_coverage_result'
	from /home/runner/work/simplecov/simplecov/lib/simplecov.rb:388:in `process_coverage_result'
	from /home/runner/work/simplecov/simplecov/lib/simplecov.rb:108:in `result'
	from /home/runner/work/simplecov/simplecov/lib/simplecov/configuration.rb:196:in `block in at_exit'
	from /home/runner/work/simplecov/simplecov/lib/simplecov.rb:215:in `run_exit_tasks!'
	from /home/runner/work/simplecov/simplecov/lib/simplecov.rb:203:in `at_exit_behavior'
	from /home/runner/work/simplecov/simplecov/lib/simplecov/defaults.rb:27:in `block in <top (required)>'
```

lib/simplecov.rb
robotdana: I was thinking of enabling it with a config like `SimpleCov.enable_for_forks` or something with documentation about how that modifies `Process.fork`, and conditionaly requiring it in SimpleCov.start. what do you think?

lib/simplecov/configuration.rb
robotdana: it gets evaluated at the end of every forked process and spams up stdout

lib/simplecov.rb
PragTob: Hm unsure. I guess I could go either way on it. In your mind, what'd be the advantage of that over telling people to do `require "simplecov/something"` ?

The upside I see is that it gives us a place to do more stuff should we need it. The downside is that it's more complex.

:man_shrugging:

lib/simplecov/configuration.rb
robotdana: My theory was making the computer do less work, but it's maybe not necessary.

lib/simplecov.rb
robotdana: nothing but that it feels like configuration

robotdana:
The new failing cuke, so 2.4 is certain it's looked at 2 files, and 2.7 is certain it's looked at 1 file, without changing anything else, which is not something i was expecting. it passes on 2.7, it passes on 2.4 if i change a 1 to a 2. I'm gonna figure that part out later today. happier with the config and cuke test otherwise?

lib/simplecov/configuration.rb
robotdana: turns out at least the SimpleCov::Formatter::HTMLFromatter spams stdout

robotdana:
reminder to myself: I can probably solve the file count issues with a filter

robotdana:
So its failing for jruby is interesting. fork isn't available. is it even relevant? should we just suppress running that cure on jruby 9? I have never used jruby before so i'm not even sure what goes into running it locally

robotdana:
Do you have a preference for how to handle the jruby issue?

lib/simplecov/configuration.rb
robotdana: yeah it looks like as it is currently setting SimpleCov.formatter to nil raises a nil.new NoMethodError

robotdana:
Thank you for the rabbits ❤️ 🐰

PR #894, opened 2020-06-19
Fix regexp for filtering out non-root results being set too early

### LINKED ISSUE #885: Multiple engines requiring each other only show their own files in the reports
Hi!

First of all, thank you all for your work! Simplecov is an awesome gem <3

Now for the downsides, I'm having some trouble configuring it in our Project. I'm working in Decidim (https://github.com/decidim/decidim), a platform to empower participatory democracy. It's actually a gem made of multiple gems (just like the `rails` gem does, it installs a bunch of gems). We're using a mono-repo approach to organize the code, each folder is a different gem. I've seen @deivid-rodriguez around, he's worked on the project too in the past 😄 

In our case, we have a set of gems that build upon a central one (`decidim-core`). This means each gem uses code from `decidim-core`. Each gem has its own `spec_helper.rb`, although they all look almost the same. We configure SimpleCov through the `.simplecov` file in our root folder. We run the test suites individually through Github Actions. Our current setup runs each gem test suite independently in different workflows (and thus in different file systems), so only one test suite is run at a time. This is effectively the same (for testing purposes) as running a single test run in your machine. We're not mergint the test suite manually. Instead, this merging is done by Codecov.

Now, whenever I run the tests for one of the gems I'd expect the report to include files from the module and from `decidim-core`. For example, running the tests from `decidim-proposals` should include `decidim-core` files, but it doesn't. I'm playing around SimpleCov configurations in https://github.com/decidim/decidim/pull/5934, but so far the reports only show files form the given module, no extra ones (even though I've set `SimpleCov.root` correctly). I've also tried the snippet in the README file to load files from engines, but without any luck.

Since we're exporting the coverage reports to Codecov, I'm using the `simplecov-cobertura` formatter, which is the one recommended in the Codecov docs. I've checked the internals of that gem and it doesn't seem to filter files, so I think the problem comes from a misconfiguration in our part.

If you'd like to check the test suite locally, these are the required steps:

1. Clone the repo and `cd` into it
1. `bundle install`
1. `bin/rake test_app`
1. `CI=1 SIMPLECOV=true bin/rspec ./decidim-proposals/spec/system/proposals_spec.rb:50`
1. Check the `coverage/coverage.xml` file

This step I'm using is a system spec that checks a `#show` page for a resource. This page uses code from `decidim-proposals` and `decidim-core`, but if you check the coverage file you'll see how it only shows code from `decidim-proposals`.

Technical details:

- Ruby 2.5.3p105
- SimpleCov 0.18.5 (latest)
- simplecov-cobertura 1.3.1 (latest)

PD: Feel free to change the issue title if you can come up with a better explanation! Also, ping me if something's not clear enough!

deivid-rodriguez:
If user configures a custom `SimpleCov.root`, results outside of the new root wouldn't be properly filtered out since the regexp for filtering was being set at require time, before the `SimpleCov.root` configuration had been applied.

Instead, lazily set this regexp, so that by the time it is used, the new `SimpleCov.root` value is already setup.

Fixes #885.

### DIFF
--- a/features/config_root.feature
+++ b/features/config_root.feature
@@ -1,0 +1,26 @@
+@rspec @disable-bundler
+
+Feature:
+
+  The root of the project can be customized.
+
+  Background:
+    Given I'm working on the project "monorepo"
+
+  Scenario: A coverage result is considered if it falls inside the root of the project
+    Given I install dependencies
+    And a file named ".simplecov" with:
+      """
+      SimpleCov.start do
+        root __dir__
+      end
+      """
+    When I open the coverage report generated with `bin/rspec_binstub_that_chdirs extra/spec/extra_spec.rb`
+    Then I should see the groups:
+      | name      | coverage | files |
+      | All Files | 100.0%   | 2     |
+
+    And I should see the source files:
+      | name                        | coverage |
+      | base/lib/monorepo/base.rb   | 100.00 % |
+      | extra/lib/monorepo/extra.rb | 100.00 % |
--- a/lib/simplecov/useless_results_remover.rb
+++ b/lib/simplecov/useless_results_remover.rb
@@ -5,12 +5,14 @@ module SimpleCov
   # Select the files that related to working scope directory of SimpleCov
   #
   module UselessResultsRemover
-    ROOT_REGX = /\A#{Regexp.escape(SimpleCov.root + File::SEPARATOR)}/io.freeze
-
     def self.call(coverage_result)
       coverage_result.select do |path, _coverage|
-        path =~ ROOT_REGX
+        path =~ root_regx
       end
+    end
+
+    def self.root_regx
+      @root_regx ||= /\A#{Regexp.escape(SimpleCov.root + File::SEPARATOR)}/i.freeze
     end
   end
 end
--- a/test_projects/monorepo/Gemfile
+++ b/test_projects/monorepo/Gemfile
@@ -1,0 +1,15 @@
+source "https://rubygems.org"
+
+gem "rspec"
+
+gem "base", path: "base"
+gem "extra", path: "extra"
+
+# when the tests are executed the project is in tmp/aruba/project
+# which is a different nesting from its usual place
+root = File.expand_path("../..", __dir__)
+unless File.exist?("#{root}/simplecov.gemspec")
+  root = File.expand_path("../../..", __dir__)
+end
+
+gem "simplecov", path: root
--- a/test_projects/monorepo/Gemfile.lock
+++ b/test_projects/monorepo/Gemfile.lock
@@ -1,0 +1,49 @@
+PATH
+  remote: ../..
+  specs:
+    simplecov (0.18.5)
+      docile (~> 1.1)
+      simplecov-html (~> 0.11)
+
+PATH
+  remote: base
+  specs:
+    base (0.0.1)
+
+PATH
+  remote: extra
+  specs:
+    extra (0.0.1)
+      base
+
+GEM
+  remote: https://rubygems.org/
+  specs:
+    diff-lcs (1.3)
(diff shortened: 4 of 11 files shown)


spec/useless_results_remover_spec.rb
```diff
@@ -32,4 +32,25 @@
     expect(subject).to have_key(source_path)
     expect(subject[source_path]["lines"]).to be_kind_of(Array)
   end
+
+  context "when changing coverage root" do
+    around do |example|
+      begin
+        previous_root_regx = SimpleCov::UselessResultsRemover.instance_variable_get(:@root_regx)
+        SimpleCov::UselessResultsRemover.instance_variable_set(:@root_regx, nil)
+        previous_root = SimpleCov.root
```
PragTob, line 41: The spec is very involved and down to nitty gritty impl details. I think it works but overall would feel better if we had a cucumber scenario covering this case. No need to add it here though, I'm fine to have it as is and might add the scenario myself after merge (if I get to it) :)

deivid-rodriguez: Yep, I agree, super implementation dependent. Another approach is to use a subprocess.

deivid-rodriguez: Had a super quick look and a cucumber scenario sounds easier and better.

PragTob: yeah I mean that's what cucumber does in our setup, so I'd just add a scenario for it later replacing this. But for getting this shipped and know that it works this is good enough as of now imo so don't feel pressured to change it up :)

PragTob: we should probably rather use a chat :laughing:

deivid-rodriguez: I can add the feature, no worries! :)

deivid-rodriguez: I gave the scenario a try, let's see how it goes :crossed_fingers:.

PragTob:
Doesn't look good to CI though:  

```
/home/runner/.rubies/ruby-2.4.10/lib/ruby/gems/2.4.0/gems/bundler-2.1.4/lib/bundler/spec_set.rb:86:in `block in materialize': Could not find rspec-support-3.9.3 in any of the sources (Bundler::GemNotFound)
	from /home/runner/.rubies/ruby-2.4.10/lib/ruby/gems/2.4.0/gems/bundler-2.1.4/lib/bundler

```

features/config_root.feature
```diff
@@ -0,0 +1,25 @@
...
+    Given I'm working on the project "monorepo"
+    And a file named ".simplecov" with:
+      """
+      SimpleCov.start do
+        root __dir__
+      end
+      """
+
+  Scenario:
+    When I open the coverage report generated with `bin/rspec_binstub_that_chdirs extra/spec/extra_spec.rb`
```
PragTob, line 17: I think you might be missing a bundle install step here

deivid-rodriguez: Yep, didn't notice locally probably because I already had everything installed. Will update.

PragTob: We're using Given I install dependencies elsewhere

PragTob: Still petitioning for a chat tool (2nd time in this PR I wrote my comment without seeing your response first)

image

deivid-rodriguez: haha, we can setup a chat tool, sure!

deivid-rodriguez: No problem about mixed replies though, it happens.

deivid-rodriguez: That did the trick! :tada:

mrcasals:
Yay! Thank you so much for taking care of this ❤️

PR #913, opened 2020-08-16
Change Result#from_hash to deal with many entries isntead one

PragTob:
Inspired by #687

This feels cleaner as we have one centralized instead of 2
things making up their own loop leaking knowledge about the
data structure. That knowledge is now more contained in the
Result class which is great.

Using select also feels much better.

If anyone was using Result.from_hash we might break their code
as it's an Array now. I hope not/nobody should :)

### DIFF
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -85,13 +85,11 @@ module SimpleCov
 
       results = result_filenames.flat_map do |filename|
         # Re-create each included instance of SimpleCov::Result from the stored run data.
-        (JSON.parse(File.read(filename)) || {}).map do |command_name, coverage|
-          SimpleCov::Result.from_hash(command_name => coverage)
-        end
+        Result.from_hash(JSON.parse(File.read(filename)) || {})
       end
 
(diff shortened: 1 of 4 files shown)


mobilutz:
@PragTob I am one of these people that use `from_hash`:
```
If anyone was using Result.from_hash we might break their code
as it's an Array now. I hope not/nobody should :)
```

Why shouldn't anyone use `from_hash`?

PragTob:
@mobilutz hi there and sorry for causing trouble for you :( I see/think of `Result.from_hash` as an internal API aka something we use but not necessary other people/gems use. That's never to easy to communicate in Ruby though ;) If I'd known I'd probably have created another method to take care of it and left Result.from_hash alone.

Can you maybe open an issue with your use case and we can talk there? Of course I want people to be able to do what they do, would just be good to know your use case and if there might be better ways.

mobilutz:
@PragTob Thanks for the response. And I guess you will see after reading #920 it was not really huge trouble for us. And it totally related to our way of using simplecov!

PR #930, opened 2020-09-17
Invalidate coverage_path cache for updating root

a2ikm:
As updating `coverage_dir`, updating `root` should invalidate `coverage_path` cache.

### DIFF
--- a/lib/simplecov/configuration.rb
+++ b/lib/simplecov/configuration.rb
@@ -22,6 +22,7 @@ module SimpleCov
     def root(root = nil)
       return @root if defined?(@root) && root.nil?
 
+      @coverage_path = nil # invalidate cache
       @root = File.expand_path(root || Dir.getwd)
     end
 


PragTob (commented):
:wave: Hi there,

thank you a lot for your contribution!

Is there a way in which you could easily show case the problems that leads to? I'm talking about a unit test or a feature test show casing this kind of problem to make sure we don't introduce regressions for that scenario.

nala

PragTob:
:wave: 

Hey there, I want to have this in but I'd also be really interested in a cucumber scenario or something to that tune to reproduce the bug so we can be safer for the future :)

a2ikm:
Thank you for your comments, and sorry for my late reply.

I’m working on a monorepo including some rails applications whose tests are run in parallel on CircleCI. So I need a script like the following steps to summarize them:

1. merge multiple .resultset.json files and dump the result to an HTML for an application
2. change the root directory for another application, and do the step 1 for it
3. repeat 1 and 2 for each application

This is why I need to invalidate the root directory.

By the way, this PR isn’t enough to do it. I wrote a ruby script with my forked simplecov gem, but it couldn’t dump appropriate data for each application. There seems other caches.

So I run two scripts for now: 

1. run the following script for each application
2. merge the JSON files and dump it

PragTob:
I see thanks. If this PR isn't enough to do it then I'd probably not merge it yet. I'd prefer if we can come up with at least one small scenario to replicate this.

In your scenario, when you say you merge it do you mean you run `SimpleCov.collate` on multiple directories or what are you doing?

a2ikm:
I think it's good not to merge this PR. More works and investigations should be done 👍 

I've missed `SimpleCov.collate`, so I used `SimpleCov::ResultMerger` manually. But, yes, I will use `SimpleCov.collate` in another chance.

a2ikm:
Then, I'd close this PR for now. Is it OK?

PragTob:
@a2ikm If you manage to investiage the problem and come up with a scenario/test to reproduce this I'd highly appreciate it. As it stand I checked the code again and you're absolutely write it should invalidate the cache... so I'll still merge it to get the fix to people It's just potentially not complete or enduring without a spec.

Thanks!

a2ikm:
Thank you for your merge!

PR #932, opened 2020-10-03
simplecov.rb - fix Process monkey patch for fork when unsupported

MSP-Greg:
Monkey patching `Process.fork` causes issues with applications that make use of fork but also run on Ruby builds without fork (like Windows & JRuby).

Fix so `./simplecov/process.rb` is only required if the platform supports `Process.fork`.

### DIFF
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -50,7 +50,9 @@ module SimpleCov
     def start(profile = nil, &block)
       require "coverage"
       initial_setup(profile, &block)
-      require_relative "./simplecov/process" if SimpleCov.enabled_for_subprocesses?
+      require_relative "./simplecov/process" if SimpleCov.enabled_for_subprocesses? &&
+                                                ::Process.respond_to?(:fork)
+
       make_parallel_tests_available
 
       @result = nil


PragTob (approved):
Thank you very much and sorry for the oversight!

I'll merge so it can go into the next release, however if you could post what the problem is that occurs I'd be happy. We're running tests on JRuby and they didn't fail so it must be a gap in our tests that'd like to fill and fix. 

WhatsApp Image 2020-07-11 at 23 18 00

MSP-Greg:
> you could post what the problem is that occurs

Sorry.  I was/am working on coverage for an app that runs on Ubuntu/macOS/Windows both MRI & JRuby.  It uses fork, and the test suite has tests for that.

Rather than using platform/OS conditionals, we've always preferred using `::Process.respond_to? :fork`, as it's clear what the intent is.  When we load simplecov, it patches Process, regardless of whether Ruby supports it.

First time I ran coverage, was a little surprised to see the fork tests running on JRuby & Windows...

JFYI, the app (Puma) has parallel tests, tests starting Puma via `IO.popen`, and the `popen` processes are running `fork`.  Not sure if it's running correctly yet.  There's coverage for the `popen` processes, but not sure about the forks.

I did need to `prepend` into `SimpleCov::Result#format!` to shut off the output, as sometimes we're testing the io returned by `IO.popen`.  I think that made sense...

Also, re CI, `stable.yml` could easily add a Windows job, and since `ruby/setup-ruby@v1` now has ruby head builds on all platforms, `unstable.yml` could use those.  Several repos test against those head builds, some not even using `allow-failure`...

PR #934, opened 2020-10-07
Implement branch coverage support for exit status modifiers

jemmaissroff:
This implements branch coverage support for `:minimum_coverage_by_file`, `:maximum_coverage_drop`, and `:refuse_coverage_drop`. 

Resolves Issue #844.

PR #953, opened 2020-12-21
Checking if CI works

PragTob:
Ref: https://github.com/simplecov-ruby/simplecov/pull/934#issuecomment-748985375

### DIFF
--- a/.test
+++ b/.test
@@ -1,0 +1,1 @@
+test test


PragTob:
Ok CI runs here... this is weird. Do I have to create a non privileged account to recreate this?

PR #955, opened 2020-12-21
Allow specs to pass against Ruby 3.0

tagliala:
Hi, I was experiencing an issue with [coveralls ruby reborn](https://github.com/tagliala/coveralls-ruby-reborn) and Ruby 3.0 that I thought could be related to a commit in SimpleCov. ~I'm still investigating, so I will let you know if there is something that needs attention here~ but I just need a change on my side because of #913

While doing that, I've noticed that feature specs didn't run with a pristine 3.0 gemset without webrick, and when they run they fail because:
- puma 4.3.1 didn't compile on recent macOS Catalina
- minitest 5.14.0 doesn't support Ruby 3.0
- Some of the gems listed in the `rspec_rails` dummy app don't work with Ruby 3.0

This commit uses a dummy 6.1 rails application and allows specs to pass on macOS Catalina 10.15.7 (19H114), as well as Ruby 3.0.0-rc1

PS: is there a reason why webrick was not listed in the development dependencies?


tagliala:
All green, but I think I can't introduce changes to GitHub actions and test against new Rubies

.github/workflows/stable.yml
```diff
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        ruby: [2.5.8, 2.6.6, 2.7.1, jruby-9.2.11.1]
+        ruby: [2.5.8, 2.6.6, 2.7.2, 3.0.0-rc1, jruby-9.2.11.1]
```
tagliala, line 11: should I remove this commit?

Gemfile
```diff
@@ -19,6 +19,7 @@ group :development do
   gem "pry"
   gem "rubocop"
   gem "test-unit"
+  gem "webrick"
```
tagliala, line 22: Is there any reason because this was missing?

test_projects/rails/rspec_rails/Gemfile
```diff
@@ -14,36 +14,46 @@ else
...
-  gem "spring"
+  gem 'spring'
 end
 
 group :test do
   # Adds support for Capybara system testing and selenium driver
-  gem "capybara", ">= 2.15"
-  gem "selenium-webdriver"
+  gem 'capybara', '>= 3.26'
+  # gem 'selenium-webdriver'
```
tagliala, line 53: not yet compatible with Ruby 3. It is not needed for the specs, so I've commented it out. Also the commit message includes the reason

.github/workflows/stable.yml
PragTob: Don't see a reason to :)

Gemfile
PragTob: webrick is part of the standard library and comes with every (normal) ruby install so we don't include it. Although, there have been discussions to include things like webrick I'm not currently a fan but still pondering the issue.

PragTob: For now, I'd prefer if you removed it

test_projects/rails/rspec_rails/Gemfile
```diff
@@ -14,36 +14,46 @@ else
 end
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem "rails", "~> 6.0.2", ">= 6.0.2.1"
+gem 'rails', '~> 6.1.0'
 # Use Puma as the app server
-gem "puma", "~> 4.1"
+gem 'puma', '~> 5.0'
```
PragTob, line 19: We use `"` in the main Gemfile so I'd prefer it stays the same here and not changed here :)

test_projects/rails/rspec_rails/Gemfile
```diff
@@ -14,36 +14,46 @@ else
...
 end
 
 group :development do
   # Access an interactive console on exception pages or by calling 'console' anywhere in the code.
-  gem "web-console", ">= 3.3.0"
+  gem 'web-console', '>= 4.1.0'
+  # Display performance information such as SQL time and flame graphs for each request in your browser.
+  # Can be configured to work on production as well see: https://github.com/MiniProfiler/rack-mini-profiler/blob/master/README.md
+  gem 'rack-mini-profiler', '~> 2.0'
+  gem 'listen', '~> 3.3'
```
PragTob, line 45: we don't need these 2, right? So I'd rather we remove them.

test_projects/rails/rspec_rails/Gemfile
PragTob: Could you please add a comment on top of it? Commit messages are nice but they also tend to get muddied and hard to track down.

PragTob (commented):
Hey, thanks for your contribution! :green_heart: :rocket: 

Looks mostly good just some smaller comments!

IMG_20200927_090520

Gemfile
tagliala: Got it. Any chance to comment this, maybe in the contributing guidelines?

PragTob: I'm not really sure - how did you even notice this? It's been like this for years and I've rarely seen anyone add stdlib gems to their Gemfiles... am I just... _old_ and that's what the cool kids do these days?

deivid-rodriguez: @PragTob The thing is, `webrick` has been removed from stdlib in ruby 3.0. So people who depend on `webrick` and were being explicit about it will not even notice the change. However, if you depended on `webrick` but were relying on `ruby` to provide that dependency, then things will break for you.

tagliala: @deivid-rodriguez thanks, the problem is with Ruby 3.0 indeed

I think I need to edit the commit message and clarify with a comment why there is a (former) stdlib in the Gemfile

PragTob: @deivid-rodriguez thanks, didn't know! Guess gotta read up on my Ruby 3. In that case @tagliala please leave it in!

PragTob: The release notes don't mention it though :thinking:  https://www.ruby-lang.org/en/news/2020/12/20/ruby-3-0-0-rc1-released/

Mentions some other gems... huh.

Here is the issue removing it though: https://bugs.ruby-lang.org/issues/17303

tagliala: I've changed the commit message and left an explicit comment on why `webrick` is there

test_projects/rails/rspec_rails/Gemfile
tagliala: Done. I've also removed all unneeded changes from this file

tagliala:
Specs fail against JRuby. Reproduced in localhost, looks like it depends on the sql adapter

```
An error occurred while loading ./spec/foo_spec.rb.
Failure/Error: ActiveRecord::Migration.maintain_test_schema!

FrozenError:
  can't modify frozen Hash
# ~/.rvm/gems/jruby-9.2.14.0/gems/activerecord-jdbc-adapter-50.0/lib/arjdbc/sqlite3/connection_methods.rb:3:in `sqlite3_connection'
```

I think it could be related to rails/rails#37291

Ref: jruby/activerecord-jdbc-adapter#1080

PragTob:
The JRuby issue seems to indicate that it should be fixed though? The version in your stacktrace is rather old/there's a a new 61.0 version. I also realized it doesn't seem to be checked in to the Gemfile, that's something that could use some fixing :)

tagliala:
@PragTob thanks for pointing me in the right direction.

I do not use JRuby, now I understand that if a dependency is listed as "java" at rubygems, it requires `platforms: %i[jruby]` to be downloaded properly

Therefore, I was getting `51.0` because of the missing `platforms` option, which I added in the dummy app Gemfile and fixed both my local build and GitHub actions.

Ref: https://rubygems.org/gems/activerecord-jdbcsqlite3-adapter/versions

Let me know if this sounds correct to you

PragTob:
I don't think it needs the `platforms` thing (because it's in the `if`) so should be fine even without. I'll check it out :)

tagliala:
> I don't think it needs the platforms thing (because it's in the if) so should be fine even without. I'll check it out :)

I was supposing the same, but that did not happen on both GitHub actions ("not able to find 61.0..." -- sorry, didn't take a screen) and on my local machine

I can push a commit without platforms so you can see the error on github actions

PragTob:
Huh weird it doesn't realize this is merged now :thinking: I probably should have rebased somewhere or rebased where I shouldn't - anyhow it's on main now so thanks a lot! :rocket: :green_heart: 

WhatsApp Image 2020-12-21 at 11 03 04

tagliala:
Welcome!

PR #956, opened 2020-12-22
Fix tests

PragTob:
#934 introduced these unknowingly because I can't figure out pull_request vs. pull_request_target in github actions


PragTob:
pull_request_target is failing because that's what's on main - ignore!

PR #960, opened 2020-12-28
Improve performance of collate & reduce memorty consumption

PragTob:
Draft to get early feedback! (cc: @deivid-rodriguez ? :D ) So Todos and fixmes are intended and will either get resolved or turned into tickets.

I'm worried about breaking things, as the work to move my [quick performance hack](https://github.com/simplecov-ruby/simplecov/compare/collate-plus-plus?expand=1) to main line code was more involved than I had hoped.

I haven't measured this again but memory consumption and speed gains were massive with the initial spike. Perf was ~2x for my smaller sample set and it allowed the big sample set to do its thing at all (I killed it at 12GB memory consumption).

------------------------------------------------------

We used to read all files into memory - which gets too much if
someone runs a massively parallel CI because 400 * 10MB is still
~4GB and that's just pure file size and we do a lot more with it.

Breaks the interface of ResultMerger.merge_and_store but it's not
intended as a public interface. Will leave a note in the Changelog
anyhow.

What is attempted to do top leve is perhaps easier to see when
looking at the spike code: https://github.com/simplecov-ruby/simplecov/compare/collate-plus-plus?expand=1

Changes go further than just not reading all files in at once,
during the merge process we also operate on the raw file structure
as opposed to creating SimpleCov::Result. Creating SimpleCov::Result
comes with a lot of overhead, notably reading in all source files. So,
that's even worse doing ~400 times in a large code base.

There's more optimization potential for cases like these which
I'll open a ticket about but notably:
* Potentially don't create SimpleCov::Result at all until we really
  produce results (just dump the raw coverage more or less)
* allow running without a formatter as only the last one really
  needs the formatter

### DIFF
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -81,17 +81,13 @@ module SimpleCov
     # information about coverage collation
     #
     def collate(result_filenames, profile = nil, &block)
-      raise "There's no reports to be merged" if result_filenames.empty?
+      raise "There are no reports to be merged" if result_filenames.empty?
 
       initial_setup(profile, &block)
 
-      results = result_filenames.flat_map do |filename|
-        # Re-create each included instance of SimpleCov::Result from the stored run data.
-        Result.from_hash(JSON.parse(File.read(filename)) || {})
-      end
-
       # Use the ResultMerger to produce a single, merged result, ready to use.
-      @result = ResultMerger.merge_and_store(*results)
+      # TODO: Did/does collate ignore old results? It probably shouldn't, right?
+      @result = ResultMerger.merge_and_store(*result_filenames)
 
       run_exit_tasks!
     end
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -63,6 +63,7 @@ module SimpleCov
 
     #
     # Collate a series of SimpleCov result files into a single SimpleCov output.
+    #
     # You can optionally specify configuration with a block:
     #   SimpleCov.collate Dir["simplecov-resultset-*/.resultset.json"]
     #    OR
@@ -86,8 +87,7 @@ module SimpleCov
       initial_setup(profile, &block)
 
       # Use the ResultMerger to produce a single, merged result, ready to use.
-      # TODO: Did/does collate ignore old results? It probably shouldn't, right?
-      @result = ResultMerger.merge_and_store(*result_filenames)
+      @result = ResultMerger.merge_and_store(*result_filenames, ignore_timeout: true)
 
       run_exit_tasks!
     end
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -81,13 +81,17 @@ module SimpleCov
     # available config options, or checkout the README for more in-depth
     # information about coverage collation
     #
-    def collate(result_filenames, profile = nil, &block)
+    # By default `collate` ignores the merge_timeout so all results of all files specified will be
+    # merged together. If you want to honor the merge_timeout then provide the keyword argument
+    # `ignore_timeout: false`.
+    #
+    def collate(result_filenames, profile = nil, ignore_timeout: true, &block)
       raise "There are no reports to be merged" if result_filenames.empty?
 
       initial_setup(profile, &block)
 
       # Use the ResultMerger to produce a single, merged result, ready to use.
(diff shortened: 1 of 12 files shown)


lib/simplecov.rb
```diff
@@ -81,17 +81,13 @@ def start(profile = nil, &block)
...
       initial_setup(profile, &block)
 
-      results = result_filenames.flat_map do |filename|
-        # Re-create each included instance of SimpleCov::Result from the stored run data.
-        Result.from_hash(JSON.parse(File.read(filename)) || {})
-      end
-
       # Use the ResultMerger to produce a single, merged result, ready to use.
-      @result = ResultMerger.merge_and_store(*results)
+      # TODO: Did/does collate ignore old results? It probably shouldn't, right?
```
PragTob, line 89: pretty sure it didn't/doesn't in the past but does right now which means we'll likely have to introduce a boolean flag to `merge_and_store`.

lib/simplecov/result.rb
```diff
@@ -85,31 +81,7 @@ def self.from_hash(hash)
 
   private
 
-    # We changed the format of the raw result data in simplecov, as people are likely
-    # to have "old" resultsets lying around (but not too old so that they're still
-    # considered we can adapt them).
-    # See https://github.com/simplecov-ruby/simplecov/pull/824#issuecomment-576049747
-    def adapt_result(result)
```
PragTob, line 92: code had to be moved since we don't create `Result` objects for merging any more

lib/simplecov/result.rb
```diff
@@ -85,31 +81,7 @@ def self.from_hash(hash)
...
-      data.is_a?(Array)
-    end
-
-    def adapt_pre_simplecov_0_18_result(result)
-      result.transform_values do |line_coverage_data|
-        {"lines" => line_coverage_data}
-      end
-    end
-
+    # This is dropping data of files that can not be found... why though?
```
PragTob, line 84: seriously... why? Can't we just let it in there and it'll be dropped on result generation anyhow as we can't display it? Probaby have to do some git digging on this to find the reason.

lib/simplecov/result_merger.rb
```diff
@@ -19,81 +19,110 @@ def resultset_writelock
...
+        end
       end
 
       #
-      # Gets all SimpleCov::Results from cache, merges them and produces a new
+      # Gets all SimpleCov::Results stored in resultset, merges them and produces a new
       # SimpleCov::Result with merged coverage data and the command_name
       # for the result consisting of a join on all source result's names
       #
+      # TODO: Maybe put synchronization just around the reading?
```
PragTob, line 103: yeah so right now it synchronizes including the whole merge time which might be a slow down - probably means extracting onr or 2 smaller methods and just puttign them together but should be worth it

lib/simplecov/result_merger.rb
```diff
@@ -116,9 +145,29 @@ def synchronize_resultset
         end
       end
 
-      # Clear out the previously cached .resultset
-      def clear_resultset
-        @resultset = nil
```
PragTob, line 121: resultset was only ever called in library code right after clearing it, so the whole clear and cache mechanism wasn't much good (unless I missed something)

spec/fixtures/conditionally_loaded_1.rb
```diff
@@ -0,0 +1,3 @@
+# some comment
+puts "wargh"
+puts "wargh 1"
```
PragTob, line 3: these files were just missing but referenced in tests and sicne we have code that removes missing files from saved results it caused problems so added them :)

lib/simplecov/result.rb
PragTob: It has existed basically forever: https://github.com/simplecov-ruby/simplecov/commit/4f37de8931b60e9731b100cd63cd0badca4ce5e3

deivid-rodriguez:
@PragTob I'm sorry I can't find time right now to review this :disappointed:

PragTob:
@deivid-rodriguez no worries at all! :green_heart: Just thought it'd be great. I'll merge it this afternoon, if at some point later you find the time I'd be happy but no pressure or worries at all. Happy new year and I hope that all is well! :rocket: 

WhatsApp Image 2020-12-10 at 09 49 49

PR #985, opened 2021-04-05
Improvements to the README

artagnon:


olleolleolle (commented):
The changes to the subprocess documentation makes use of the `start` block "DSL", which helps the first-time user see those methods in use. I'm guessing that the docs mention the methods being available in "the class method way", too.

I'm in favor of improving the documentation to make it more scannable. This diff is big, though.

Perhaps 1 PR could be the "Markdown formatting changes" and another the "explanation improvements"?  Formatting changes were intermingled with the initial changes to the subprocess code example, which made the diff much bigger.

artagnon:
Thanks; let me open another PR.

Issue #782, opened 2019-12-05
Implement support for method coverage


PragTob:
Ruby 2.5 didn't just add branch coverage but also method coverage (see here for instance: https://blog.bigbinary.com/2018/04/11/ruby-2-5-supports-measuring-branch-and-method-coverages.html)

Would be nice to have some support for it as well.

cc: @tycooon (you mentioned you had it implemented in your fork already)

tycooon:
Yes, I am going to work on creating PR with that functionality.

PragTob:
@tycooon that'd be amazing, but don't feel pressured. Just mentioned you since you said you had it :)

klyonrad:
That's another thing where you/we might discuss on how to actually *show* the coverage of methods.

Currently: when a file is loaded, all the `def method_name` lines are considered as a hit, the actual code in the 

and the closing of the method is also not considered as a hit
(from the article)
> Coverage was disabled for line number 7 as it contains just end keyword.

It always itched me that just declaring the method names are recorded as a hit or miss (except when nothing loads the file, which is the default for a rails project).

Anyway, that's maybe a distraction from the original topic...

**What exactly should be shown for the method coverage?** How does it affect the total big number of coverage a file/the project (that metric that we all care about). How do dynamically defined methods work with this? Playing devil's advocate: What does it even matter - I mean, what decisions would made knowing that a file has x amount of uncovered methods?

tycooon:
In my experience, method coverage mostly helps finding unused methods that can be safely deleted in case you have 100% line (and probably branch) coverage.

For example you might have `attr_reader :foo, :bar` where only `:foo` is actually used. Or simply `attr_reader :bar` that is never used. In both cases only method coverage can find such stuff.

klyonrad:
Wait a second, you can detect unused `attr_reader` methods? That's amazing! Thanks for clearing that up.

Then I would say that for an unused method the `def method_name` line should also stay red - even when the `LineCoverage` API reports it as covered

tycooon:
Well, I guess it's possible to do that, but maybe we can somehow separate the presentation of different coverage types so that they don't interfere.

For example, in HTML report we could add some extra styling for method names that are not covered. Not sure about that though 🤔

klyonrad:
But that would make it complicated for the users. Why should they care if it is `LineCoverage` or `MethodCoverage`. It doesn't matter. **The method is not called, so it is not covered.**

[...]

PR #987, opened 2021-04-23
Add method coverage support

tycooon:
See #782.

- I had to rework how coverage is parsed/unparsed (see `ResultSerialization` module).
- Because of that, I had to revert some of `ResultMerger` optimizations :(
- Coverage is now always stored in memory in symbolized form (`lines: []` instead of `"lines" => []`), just the same way you get it when calling Ruby's `Coverage.result`. Because of that, a lot of specs were updated.
- Sometimes method owner includes memory address (e.g. when method is called on anonymous class). This makes results merging problematic, so we replace memory addresses with zero address `0x0000000000000000`.
- Removed `adapt_pre_simplecov_0_18_result` stuff, since it's problematic to support this stuff and anyway the 0.18 version is already kind of outdated.

All this stuff has been battle-tested on a couple of big projects, some of which merge results of multiple CI jobs.


tycooon:
@PragTob hope you will find some time to review the changes.

lib/simplecov/source_file.rb
```diff
@@ -221,9 +242,9 @@ def ensure_remove_undefs(file_lines)
     end
 
     def build_lines
-      coverage_exceeding_source_warn if coverage_data["lines"].size > src.size
+      coverage_exceeding_source_warn if lines_data.size > src.size
```
dan-jensen, line 245: In `0.21.2` I'm encountering an error on this line when executing via [rubycritic](https://rubygems.org/gems/rubycritic).  Turns out `coverage_data` is an Array ([as documented](https://github.com/simplecov-ruby/simplecov/pull/987/files#diff-81a03b411bc0f94f1d0d8c6ae5065779a9ec8aabb88c9a15f7f8c5f47811a3faR11)), and [was previously an Integer](https://github.com/simplecov-ruby/simplecov/commit/6e7a9ab3fa2c03ee86930596139270d822e2cc77), so did this line ever work?  Related to that, how does [`coverage_data.fetch(:branches, {})` (below)](https://github.com/simplecov-ruby/simplecov/pull/987/files#diff-81a03b411bc0f94f1d0d8c6ae5065779a9ec8aabb88c9a15f7f8c5f47811a3faR293) work?

tycooon: `coverage_data` is a Hash in this PR. So it's strange that you have problems. Maybe you have coverage data in some old format that is breaking the code?

tycooon: Update: I see that rubycritic is doing [a lot of stuff](https://github.com/whitesmith/rubycritic/blob/main/lib/rubycritic/analysers/coverage.rb) with simplecov internals. So I guess they will have to update the code if this PR is merged. They already do something similar [here](https://github.com/whitesmith/rubycritic/blob/main/lib/rubycritic/analysers/coverage.rb#L112) for example.

dan-jensen: @tycooon just to be clear, I'm not having any problem with this PR.  I think this PR will solve my problem 🙂 

Also for clarity: the error I encountered was `no implicit conversion of String into Integer (TypeError)`.  From my reading of the traceback it looks like this error is internal to `simplecov`.  Yes, it's possible that it's due to some sort of mishandling by `rubycritic` (which invokes `simplecov` through [`find_coverage_percentage`](https://github.com/whitesmith/rubycritic/blob/main/lib/rubycritic/analysers/coverage.rb#L33)).  But it seems smart to focus on `simplecov` first because the existing code doesn't seem to make sense (which is what I was describing above).  I'm looking forward to this PR helping to clarify `SourceFile`, and hopefully also fixing the bug.

<details>
<summary>Traceback</summary>
running simple_cov
Traceback (most recent call last):
	19: from {PATH}bin/rubycritic:23:in `<main>'
	18: from {PATH}bin/rubycritic:23:in `load'
	17: from {PATH}lib/ruby/gems/2.7.0/gems/rubycritic-4.6.1/bin/rubycritic:10:in `<top (required)>'
	16: from {PATH}lib/ruby/gems/2.7.0/gems/rubycritic-4.6.1/lib/rubycritic/cli/application.rb:21:in `execute'
	15: from {PATH}lib/ruby/gems/2.7.0/gems/rubycritic-4.6.1/lib/rubycritic/commands/default.rb:19:in `execute'
	14: from {PATH}lib/ruby/gems/2.7.0/gems/rubycritic-4.6.1/lib/rubycritic/commands/default.rb:24:in `critique'
	13: from {PATH}lib/ruby/gems/2.7.0/gems/rubycritic-4.6.1/lib/rubycritic/analysers_runner.rb:29:in `run'
	12: from {PATH}lib/ruby/gems/2.7.0/gems/rubycritic-4.6.1/lib/rubycritic/analysers_runner.rb:29:in `each'
	11: from {PATH}lib/ruby/gems/2.7.0/gems/rubycritic-4.6.1/lib/rubycritic/analysers_runner.rb:32:in `block in run'
	10: from {PATH}lib/ruby/gems/2.7.0/gems/rubycritic-4.6.1/lib/rubycritic/analysers/coverage.rb:20:in `run'
	 9: from {PATH}lib/ruby/gems/2.7.0/gems/rubycritic-4.6.1/lib/rubycritic/core/analysed_modules_collection.rb:32:in `each'
	 8: from {PATH}lib/ruby/gems/2.7.0/gems/rubycritic-4.6.1/lib/rubycritic/core/analysed_modules_collection.rb:32:in `each'
	 7: from {PATH}lib/ruby/gems/2.7.0/gems/rubycritic-4.6.1/lib/rubycritic/analysers/coverage.rb:21:in `block in run'
	 6: from {PATH}lib/ruby/gems/2.7.0/gems/rubycritic-4.6.1/lib/rubycritic/analysers/coverage.rb:38:in `find_coverage_percentage'
	 5: from {PATH}lib/ruby/gems/2.7.0/gems/simplecov-0.21.2/lib/simplecov/source_file.rb:81:in `covered_percent'
	 4: from {PATH}lib/ruby/gems/2.7.0/gems/simplecov-0.21.2/lib/simplecov/source_file.rb:35:in `coverage_statistics'
	 3: from {PATH}lib/ruby/gems/2.7.0/gems/simplecov-0.21.2/lib/simplecov/source_file.rb:333:in `line_coverage_statistics'
	 2: from {PATH}lib/ruby/gems/2.7.0/gems/simplecov-0.21.2/lib/simplecov/source_file.rb:241:in `lines_strength'
	 1: from {PATH}lib/ruby/gems/2.7.0/gems/simplecov-0.21.2/lib/simplecov/source_file.rb:43:in `lines'
{PATH}lib/ruby/gems/2.7.0/gems/simplecov-0.21.2/lib/simplecov/source_file.rb:224:in `build_lines': no implicit conversion of String into Integer (TypeError)
</details>

0exp:
@PragTob ping!

spec/fixtures/methods.rb
gondalez, line 1: Thanks for your work on this PR, it would make a welcome inclusion 🙏 

I am curious - would method coverage improve coverage support for ruby 3's [endless methods](https://bugs.ruby-lang.org/issues/16746)?

I have found that endless methods currently always show as covered, even if they are never executed by tests.
Even if you have branching coverage enabled they report as fully covered.

Example class, spec and coverage report follow.
Note that `uncovered_endless_method` shows as covered despite not being executed by the test.

``` ruby
class CoveragePoc
  # this method is fully covered
  def self.simple_method
    :simple
  end

  # the :b branch for these methods is not covered
  def self.endless_branching_method(arg) = arg ? :a : :b

  def self.branching_method(arg)
    arg ? :a : :b
  end

  # these methods are not covered at all
  def self.uncovered_endless_method = :uncovered_endless

  def self.uncovered_method
    :uncovered
  end
end
```

``` ruby
RSpec.describe CoveragePoc do
  it 'works' do
    expect(described_class.simple_method).to eq(:simple)
    expect(described_class.branching_method(true)).to eq(:a)
    expect(described_class.endless_branching_method(true)).to eq(:a)
  end
end
```

CleanShot 2024-01-04 at 13 28 26@2x

amatsuda:
@tycooon Thank you for working on this! Yes, I'm very much interested in this feature, and willing to take a look once the patch is cleanly rebased against the main branch :)

lib/simplecov/source_file.rb
```diff
@@ -154,6 +155,26 @@ def line_with_missed_branch?(line_number)
       branches_for_line(line_number).any? { |_type, count| count.zero? }
     end
 
+    def methods
+      @methods ||= build_methods
+    end
+
+    def total_methods
+      @total_methods ||= covered_methods + missed_methods
+    end
```
sferik: The `total_methods` method defined in `FileList` returns a count, as the name suggests, but this one returns an array. Please change the name of this one (or change its behavior to be consistent with the other).

lib/simplecov/source_file/method.rb
```diff
@@ -0,0 +1,37 @@
...
+      def missed?
+        !skipped? && coverage.zero?
+      end
+
+      def lines
+        @lines ||= source_file.lines[(start_line - 1)..(end_line - 1)]
+      end
+
+      def to_s
+        "#{klass}##{method}"
```
sferik: Ruby’s `Coverage.result` uses different prefixes for instance methods vs class methods (e.g., `#<Class:Foo>` for singleton methods). The `#` separator in `to_s` could be misleading for class methods. Consider using `.` for class methods.

PR #1005, opened 2021-11-23
Add needed require line for CSV example

pamelafox:
We got confused when trying to use the CSVFormatter and discovered it also required another gem (at least for our version). Thus I suggest adding a clarification as I suggest in this PR. It might also be more clear if the "Multiple formatters" heading was below the "JSON formatter" section.


PragTob:
Thanks and sorry for the confusion!
IMG_20220101_140059_Bokeh

PR #1022, opened 2022-07-07
Support parallel_tests with multiple commands

BenoitMC:
Currently, when using parallel_tests with 2 different test frameworks (eg RSpec + Cucumber), the guess is the same for the both commands and the second command results overwrite the first command results.

This PR change the guess to something like `RSpec (1/2)` instead of just `(1/2)` to have different command names and merge the results correctly.


pboling:
I hope this can be merged!!

PR #1034, opened 2022-11-11
[Security] Bumped Rack version to 2.2.3.1

nhtruong:
Rack 2.2.3 suffers from 2 CVEs which are fixed in 2.2.3.1: 
https://github.com/rack/rack/blob/main/CHANGELOG.md#2231---2022-05-27

Signed-off-by: Theo Truong <theotr@amazon.com>

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -115,7 +115,7 @@ GEM
     public_suffix (4.0.6)
     racc (1.6.0)
     racc (1.6.0-java)
-    rack (2.2.3)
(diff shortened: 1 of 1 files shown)


colszowka:
Thanks! Although I think we shouldn't actually have a Gemfile.lock inside the repo for a gem, but my naive attempt at removing it in https://github.com/simplecov-ruby/simplecov/pull/1040 didn't pass so we'll revisit that later

Issue #1020, opened 2022-05-30
Sub-process coverage is not working for ruby >= 3.1


maxat-mansurov-jetbrains:
### Setup
- OS: MacOS Monterey 12.4
- VM: rbenv
- Ruby: ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-darwin21]
- Gems:
  - Bundler: 2.3.11
  - Simplecov: 0.21.2
  - Minitest: 5.15.0

### Problem description
It seems that on rubies 3.1 & 3.2 the subprocess coverage in SimpleCov has stopped working. When we start the `SimpleCov` as following:
```
SimpleCov.start do
  enable_for_subprocesses(true)
end
```
and test code that uses the `Process.fork`, we always get the error "**coverage measurement is already setup (RuntimeError)**". In older versions of ruby (<= 3.0) the above setup worked without a problem. Presumably this is related to https://github.com/simplecov-ruby/simplecov/issues/1003. Could somebody please take a look? Thank you!

### How to reproduce

1. Download the files from this gist: https://gist.github.com/maxat-mansurov-jetbrains/1fb1a5c5746c6ce3aefa70ab5ddb7f37
2. Run the test with `$ bundle exec ruby test_fork_coverage.rb`
3. Observe the error in the output:
```
$ bundle exec ruby test_fork_coverage.rb 
Run options: --seed 7638

# Running:

/Users/maxat.mansurov/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/simplecov-0.21.2/lib/simplecov.rb:354:in `start': coverage measurement is already setup (RuntimeError)
        from /Users/maxat.mansurov/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/simplecov-0.21.2/lib/simplecov.rb:354:in `start_coverage_with_criteria'
        from /Users/maxat.mansurov/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/simplecov-0.21.2/lib/simplecov.rb:343:in `start_coverage_measurement'
        from /Users/maxat.mansurov/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/simplecov-0.21.2/lib/simplecov.rb:59:in `start'
[...]

PR #1035, opened 2022-11-28
Support Ruby 3.0, 3.1 and 3.2

mame:
This PR has two commits to make the test suite pass on Ruby 3.0, 3.1 and 3.2

* Fix https://github.com/simplecov-ruby/simplecov/issues/1020
* Update gems of test_projects/rails/rspec_rails/Gemfile


mame:
CI passed.

Ruby 2.5 had to be removed from CI due to test code. The library code itself works.
Also, the code to check if branch coverage is supported or not did not work well on JRuby 9.3. So I tweaked it.

Can you review this PR?

colszowka:
@mame Thank you! :clap: 

I think since Ruby 2.6 is also already EOL it would be fine to actually also drop that as well from CI and official support, and then adjust the [corresponding section in the README](https://github.com/simplecov-ruby/simplecov#ruby-version-compatibility) to request Ruby 2.7 or JRuby 9.3. Would you be interested in updating the PR accordingly so we have it all in one place? Otherwise I can also merge and will make the corresponding changes as a followup

mame:
@colszowka Hi, thank you for your quick response! I have added a commit to drop 2.6 and update README.md as you said. If you have any other concerns, please feel free to let me know.

colszowka:
@mame Thank you, for taking the time to adjust those things! Merged :heavy_check_mark:

bmancini42:
Excited to see this! We are upgrading our Rails host to Ruby 3 / Rails 7 and I was concerned we might need to discontinue use of `simplecov`. Any idea how soon this will be released / available as a gem?

lib/simplecov.rb
```diff
@@ -342,7 +342,7 @@ def start_coverage_measurement
       if coverage_start_arguments_supported?
         start_coverage_with_criteria
       else
-        Coverage.start
+        Coverage.start unless Coverage.running?
```
notEthan, line 345: just to note, `Coverage.running?` is not implemented on Truffleruby, so this fails. I have opened an issue there: https://github.com/oracle/truffleruby/issues/2813

eregon: Are these `unless Coverage.running?` checks necessary?
Otherwise I would suggest to drop them to avoid breaking existing TruffleRuby releases and potentially many CI runs on TruffleRuby.

Of course we'll implement the method on TruffleRuby master, but unlikely before 2023 and won't affect existing releases.

PragTob: Looks like the issue is partially described in https://github.com/simplecov-ruby/simplecov/pull/1035/commits/1cc7650454fc7c53e7a724941662c3e9d3329ccc

`Coverage.start` raises now if already running.

It also feels like something would be wrong if you started coverage while it's already running. I have no doubt though that maybe the SimpleCov code, but most definitely various degrees of client code would break.

Since we are (still :P) 0.x and "well actually" Ruby 3 broke compatibility  I'm kinda ok with that personally.

@mame am I missing something? :thinking: 

Also, hi all :wave: Been a long time, wanted to get back into this sooner but first stress, then covid hit, then bunny problems hit. Slowly but surely getting here.

eregon: Thanks for the context. I think then the best would be to add a private method like `coverage_enabled?` and track it as an ivar on `SimpleCov` and use `Coverage.running?` if available and that ivar otherwise. I'll try to make a PR.

notEthan: I opened a PR somewhat like that: https://github.com/simplecov-ruby/simplecov/pull/1043 - I'm not sure the ivar is needed

I hope your bunny(ies) are well, @PragTob.

PR #1043, opened 2022-12-25
allow Coverage to start when Coverage.running? is not implemented

notEthan:
Amends https://github.com/simplecov-ruby/simplecov/pull/1035/commits/1cc7650454fc7c53e7a724941662c3e9d3329ccc from #1035

Another option would be to just call Coverage.start unconditionally and rescue the RuntimeError it now raises when already started.

### DIFF
--- a/lib/simplecov.rb
+++ b/lib/simplecov.rb
@@ -342,7 +342,7 @@ module SimpleCov
       if coverage_start_arguments_supported?
         start_coverage_with_criteria
       else
-        Coverage.start unless Coverage.running?
+        Coverage.start unless coverage_running?
       end
     end
 
@@ -353,7 +353,13 @@ module SimpleCov
 
       start_arguments[:eval] = true if coverage_for_eval_enabled?
 
-      Coverage.start(start_arguments) unless Coverage.running?
+      Coverage.start(start_arguments) unless coverage_running?
+    end
+
+    def coverage_running?
+      # for ruby versions which do not implement Coverage.running?,
+      # Coverage.start may be called multiple times without raising.
+      Coverage.respond_to?(:running?) && Coverage.running?
     end
 
     CRITERION_TO_RUBY_COVERAGE = {


eregon (approved):
This looks good to me, thank you!

lib/simplecov.rb
```diff
@@ -353,7 +353,15 @@ def start_coverage_with_criteria
 
       start_arguments[:eval] = true if coverage_for_eval_enabled?
 
-      Coverage.start(start_arguments) unless Coverage.running?
+      Coverage.start(start_arguments) unless coverage_running?
+    end
+
+    def coverage_running?
+      Coverage.running?
+    rescue NoMethodError
```
eregon, line 361: `Coverage.respond_to? :running?` seems nice than `rescue NoMethodError`, notably because it avoids an exception in case it's not available (and doesn't catch other issues potentially).

eregon:
FYI I'm planning to try adding TruffleRuby in CI to catch such issues faster, but that needs more work so I would suggest merging this first.

eregon:
Confirmed this PR reduces `bundle exec rake spec` from 9 to just 1 failure on truffleruby-dev.

lib/simplecov.rb
notEthan: my thought was that the much more common case will have Coverage.running? implemented so checking for that every time would be slower than rescuing in the rarer case. but this is not a place where slowness matters at all, and I do prefer not to use exceptions for control flow, so agree with you and have updated it.

notEthan:
I also started looking yesterday at adding truffle to the CI here, but got hung up on aruba/childprocess, then saw that you were ahead of me and had opened an issue about the same already.

eregon:
@PragTob @colszowka Hello, it would be great if this can be merged and released when you have some open-source time, because the latest release breaks a bunch of CIs testing TruffleRuby (e.g. see backlinks to this PR)

notEthan:
Yes, it would be quite helpful if this fix could make it into a release.

eregon:
Ping, please merge and release this. It's been one month and a half and it still breaks the CI of e.g. https://github.com/ronin-rb/ronin/actions/runs/4089986844/jobs/7053088486
Can I do anything to help with this?

AlexWayfer:
I've been stuck with this: https://github.com/filewatcher/filewatcher-matrix/pull/47

PragTob:
Thanks for this, sorry for all the silence. Life and more happend, been a tough year+. I'll merge, it should be fine _famous simplecov maintainer last words_

Thanks y'all for your work :green_heart: 

@notEthan : Very cute bunny avatar!

notEthan:
Thanks for merging. A release with this would make it usable.

eregon:
@notEthan If the issue is SimpleCov 0.21.2 + TruffleRuby, that's solved in recent TruffleRuby releases and since https://github.com/simplecov-ruby/simplecov/issues/1044#issuecomment-1529783628, because TruffleRuby has `Coverage.running?` since then.

PR #1048, opened 2023-01-02
Added dependabot for GitHub Actions and bundler

ydah:
How about using dependabot in this way?
When this works, a PullRequest is created as follows:
- ruby/bigdecimal#242

It was created with reference to the following:
- ruby/csv#265


amatsuda:
@ydah Thanks!

PR #1052, opened 2023-02-01
Fix stdlib coverage 404 link

zzak:
We can link to the official Ruby docs for the gem since it's still packaged with Ruby.


zzak:
OMG, sorry. Duplicate of #1049 and #1045 🤦

PR #1053, opened 2023-02-21
Fix for Coverage.running? and add TruffleRuby in CI

### LINKED ISSUE #1044: Please add TruffleRuby to your CI
Recently simplecov-0.22.0 which uses [Coverage.running?](https://my.diffend.io/gems/simplecov/0.21.2/0.22.0). Unfortunately, [truffleruby does not support `Coverage.running?`](https://github.com/oracle/truffleruby/issues/2813). This broke the CI of many of my gems which I both use simplecov and test against truffleruby (stable). I feel like this could have been avoided by adding truffleruby to simplecov's own CI, which would have caught the issue early, and allowed time for the TruffleRuby team to add `Coverage.running?`.

eregon:
This is the fix of https://github.com/simplecov-ruby/simplecov/pull/1043 + adding TruffleRuby CI. Fixes #1044.

There is one test failing (`testunit_bad.rb`), I'm not sure why yet, but it seems valuable to run all other tests on TruffleRuby until that's figured out.

I also simplified both workflows, especially `.github/workflows/unstable.yml` was a lot more complicated than necessary.


.github/workflows/stable.yml
```diff
@@ -29,13 +30,10 @@ jobs:
...
-        run: |
-          bundle config set --local without benchmark
-          bundle install --jobs=3
-
       - name: Run tests
         run: bundle exec rake
+        if: matrix.ruby != 'truffleruby'
+
+      - name: Run specs (truffleruby)
+        run: bundle exec rake spec
```
eregon, line 38: I'm not running `cucumber` on TruffleRuby because that doesn't work due to old cucumber using old aruba using childprocess which is not really maintained currently (https://github.com/enkessler/childprocess/issues/172).

eregon:
CI is green.
@colszowka Could you review and merge this?

AlexWayfer:
We should decrease bus-factor on such repositories.

@amatsuda, sorry, are you available? You seem to be a member of @simplecov-ruby

Issue #1016, opened 2022-02-01
test_frameworks profile hides source directories with names like 'spec' or 'features'


dmolesUC:
## Summary

I'm the author of a gem called [ruby-marc-spec](https://github.com/BerkeleyLibrary/marc-spec) that for domain-nomenclature reasons, has a main module called `MARC::Spec`, with source in a corresponding directory `lib/marc/spec`.

I recently noticed that SimpleCov was only covering the root file, `lib/marc/spec.rb`, and ignoring all the files under `lib/marc/spec`.

## Steps to reproduce

1. Clone the linked repo and check out [tag `0.1.1`](https://github.com/BerkeleyLibrary/marc-spec/tree/0.1.1/).
2. In the project root, run `bundle install`.
3. Run `bundle exec rake coverage`.

### Expected

- Coverage task fails with `Coverage (99.87%) is below the expected minimum coverage (100.00%).`
- Generated HTML report in `artifacts/rcov` includes all source files under `lib`, including `lib/marc/spec/queries/query_executor.rb`, which has an uncovered line 47.

### Actual

- Coverage task suceeds.
- Generated HTML report shows only `lib/marc/spec.rb`, with 100% coverage.

## Discussion

The issue is with the filters in [`test_frameworks.rb`](https://github.com/simplecov-ruby/simplecov/blob/v0.21.2/lib/simplecov/profiles/test_frameworks.rb), plain strings which each produce a `StringFilter` that will check for the presence of the substring anywhere in the file path. A regex filter, e.g. `%r{^/spec/}`, would instead exclude these directories only at the root of the project, which seems to be the intent.

(I also wonder about the similar [`bundler_filter.rb`](https://github.com/simplecov-ruby/simplecov/blob/v0.21.2/lib/simplecov/profiles/bundler_filter.rb), but that one seems less likely to produce accidental false positives.)

## Workarounds

In my case, the `test_frameworks` profile was being pulled in by the `rails` profile, which I didn't need (lazy cut-and-paste from a Rails project). I dropped back to the default profile and added a custom regex filter for the `/spec` directory to my `.simplecov`:

```rb
require 'simplecov-rcov'

SimpleCov.start do
  add_filter %r{^/spec/}
  add_filter 'module_info.rb'

  coverage_dir 'artifacts'
  formatter SimpleCov::Formatter::RcovFormatter
  minimum_coverage 100
end
```

[...]

PR #1062, opened 2023-07-05
README: Recommend using Regexp filters for `add_group`

splattael:
Using String filters can include wrong files to groups because it matches if the string is contained within the path.

For example:

```ruby
  # This matches files like `./lib/foo.rb`
  # but also `./app/models/library.rb'
  add_group 'Libraries', 'lib'

  # Prefer Regexp filter instead:
  add_group 'Libraries', %r{^/lib/}
```

See also https://gitlab.com/gitlab-org/gitlab/-/issues/416910

### DIFF
--- a/README.md
+++ b/README.md
@@ -489,12 +489,12 @@ Add your groups with:
 
 ```ruby
 SimpleCov.start do
-  add_group "Models", "app/models"
-  add_group "Controllers", "app/controllers"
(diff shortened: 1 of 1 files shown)


splattael:
This is somewhat related to https://github.com/simplecov-ruby/simplecov/issues/1016.

:wave: @amatsuda If you have, do you mind reviewing this MR? :pray:

PR #1068, opened 2023-08-28
Support nocov comment at the end of a line

cyberdelia:
Support `nocov` comments at the end of a line to mark it as irrelevant, it's a bit less verbose than blocks to ignore some benign things, i.e:

```ruby
case value
when String
  StringFilter
when Array
  ArrayFilter
else
  raise "Absurd case" # :nocov: 
end
```

### DIFF
--- a/lib/simplecov/lines_classifier.rb
+++ b/lib/simplecov/lines_classifier.rb
@@ -12,12 +12,23 @@ module SimpleCov
     COMMENT_LINE = /^\s*#/.freeze
     WHITESPACE_OR_COMMENT_LINE = Regexp.union(WHITESPACE_LINE, COMMENT_LINE)
 
-    def self.no_cov_line
+    def self.no_cov_block
       /^(\s*)#(\s*)(:#{SimpleCov.nocov_token}:)/o
+    end
+
+    def self.no_cov_line
+      /#(\s*)(:#{SimpleCov.nocov_token}:)(\s*)$/o
     end
 
     def self.no_cov_line?(line)
       no_cov_line.match?(line)
+    rescue ArgumentError
+      # E.g., line contains an invalid byte sequence in UTF-8
+      false
+    end
+
+    def self.no_cov_block?(line)
+      no_cov_block.match?(line)
     rescue ArgumentError
       # E.g., line contains an invalid byte sequence in UTF-8
       false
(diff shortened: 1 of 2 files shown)


cyberdelia:
cc/ @amatsuda

cyberdelia:
Rebased.

cyberdelia:
Freshly rebased.

cyberdelia:
Freshly rebased!

cc/ @PragTob

will:
@amatsuda please consider this pull request

sferik (changes requested):
I have a bit of feedback but support this feature overall.

Somewhat related: as we approach a 1.0.0 release, I was considering changing the block syntax to be more explicit. Something like:

```ruby
case coverage_criterion
when :line   then compute_line
when :branch then compute_branch
when :method then compute_method
else
  # coverage:disable
  raise "Unknown criterion: #{coverage_criterion}"
  # coverage:enable
end
```

If that was the block syntax, presumably we’d want the inline syntax to match:

```ruby
case coverage_criterion
when :line   then compute_line
when :branch then compute_branch
when :method then compute_method
else
  raise "Unknown criterion: #{coverage_criterion}" # coverage:disable
end
```

This seems as good a place as any to debate making this change.

PragTob:
@sferik yeah, changes live on in the API and for future support so I believe it's worth considering/talking about.

I don't think we need different sub types of coverage disable/enable. The most common use case for coverage disable is "these lines of code do not matter" and I don't think the type of coverage makes a difference there. I should have been more explicit, but the question of how does it play with other types of coverage is mote like: 

I'm ignoring one line of code inside a branch - will the branch be counted? I'm ignoring all but one line of code inside a branch - will it be counted? I'm ignoring the condition that causes a branch, will the branch be counted?

I personally think it's fine to say: Undefined behavior, you did this to yourself. If I'd wish for anything, as long as any line inside a branch is not ignored, the branch shouldn't be ignored. But I'm not sure how much code would be needed to support that or in fact not sure what the current behavior is. 

But people will complain that it works different from what they imagine :)

sferik:
I just chatted with @PragTob and we decided on the following comment syntax:

```ruby
# simplecov:disable line
# simplecov:disable method
# simplecov:disable branch
# simplecov:disable method,branch         <- disables method and branch coverage
# simplecov:disable                       <- disables all three
# simplecov:enable                        <- enables all three
raise "Absurd case" # simplecov:disable   <- just disables this line, no need to re-enable
```

sferik:
Resolved by https://github.com/simplecov-ruby/simplecov/pull/1179.

PR #1069, opened 2023-08-29
Replace Kernel.open with IO.popen

cyberdelia:
Replace `Kernel.open` with `IO.popen` since it's deprecated and will be removed in [future ruby versions](https://github.com/simplecov-ruby/simplecov/actions/runs/6003574534/job/16282314756?pr=1068).


cyberdelia:
Seems to have been fixed in f2e02de098864627de16d2fd154f06b56ce17738 ¯\_(ツ)_/¯

PR #1079, opened 2023-12-01
Add TruffleRuby in CI

eregon:
Fixes https://github.com/simplecov-ruby/simplecov/issues/1044.

@amatsuda Could you review and merge this?

### DIFF
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -33,6 +33,9 @@ jobs:
         run: bundle exec rake
         if: matrix.ruby-version != 'truffleruby'
 
+      # Run only `rake spec` on truffleruby, because just `rake` runs `rubocop spec cucumber`.
+      # There is no value to rerun rubocop here.
+      # cucumber fails because it uses an old childprocess which depends on fork.
       - name: Run specs (truffleruby)
         run: bundle exec rake spec
         if: matrix.ruby-version == 'truffleruby'
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -19,6 +19,7 @@ jobs:
           - '3.2'
           - '3.3'
           - jruby-9.4
+          - truffleruby
 
     steps:
       - uses: actions/checkout@v4
@@ -28,5 +29,10 @@ jobs:
           ruby-version: ${{ matrix.ruby-version }}
           bundler-cache: true
 
-      - run: |
-          bundle exec rake
+      - name: Run tests
+        run: bundle exec rake
+        if: matrix.ruby-version != 'truffleruby'
+
+      - name: Run specs (truffleruby)
+        run: bundle exec rake spec
+        if: matrix.ruby-version == 'truffleruby'
--- a/.github/workflows/unstable.yml
+++ b/.github/workflows/unstable.yml
@@ -26,12 +26,10 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby-version }}
           bundler-cache: true
-        continue-on-error: ${{ (matrix.ruby-version == 'ruby-head') || (matrix.ruby-version == 'jruby-head') }}
 
       - name: Run tests
         run: bundle exec rake
         if: matrix.ruby-version != 'truffleruby-head'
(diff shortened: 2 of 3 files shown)


.github/workflows/stable.yml
```diff
@@ -26,3 +27,8 @@ jobs:
 
       - name: Run tests
         run: bundle exec rake
+        if: matrix.ruby != 'truffleruby'
+
+      - name: Run specs (truffleruby)
+        run: bundle exec rake spec
```
eregon, line 33: The reason is `bundle exec rake` runs `rubocop spec cucumber`.
`rubocop` is pointless to run again here.
`cucumber` fails because it tries to use `fork` (most likely `childprocess` tries to).
IIRC that's fixed in latest cucumber/aruba, but updating them seems out of scope of this PR.

PragTob: Agree on rubocop but that's for another day to fix/make this faster

.github/workflows/unstable.yml
```diff
@@ -27,6 +28,11 @@ jobs:
...
-      - run: |
-          bundle exec rake
+      - name: Run tests
+        run: bundle exec rake
+        if: matrix.ruby-version != 'truffleruby-head'
         continue-on-error: ${{ (matrix.ruby-version == 'ruby-head') || (matrix.ruby-version == 'jruby-head') }}
+
+      - name: Run specs (truffleruby)
+        run: bundle exec rake spec
+        if: matrix.ruby-version == 'truffleruby-head'
```
PragTob, line 38: This should also continue-on-error right? We don't want the test suite to break

spec/return_codes_spec.rb
```diff
@@ -34,6 +34,7 @@
         end
 
         it "prints a message to STDERR" do
+          skip "fails on truffleruby" if RUBY_ENGINE == "truffleruby" && command.include?("testunit_bad.rb")
```
PragTob, line 37: @eregon could you specify why it fails on TruffleRuby, what's broken, where is the bug tracked if it is tracked?

.github/workflows/unstable.yml
eregon: This workflow is a daily, so it won't be shown on PRs or even on commits, only at https://github.com/simplecov-ruby/simplecov/actions/workflows/unstable.yml.
So I think there is little value to hide the red icon if it's actually failing.

eregon: Ah but the other steps have `continue-on-error: ${{ (matrix.ruby-version == 'ruby-head') || (matrix.ruby-version == 'jruby-head') }}`, I see.
I think it's best to remove those, or simplify to `continue-on-error: true` then.

PragTob: probably continue-on-error: true yeah.

I recently took a deep dive on it and yeah it prevents the red from showing up... that saying if it's just daily I think we can display it, the badge in the README shouldn't be affected so maybe you are right and we can remove them.

eregon: I'll push a commit to remove the `continue-on-error` lines then, it seems clearer that way with no practical downside. Can always be changed if we find a downside to it.

.github/workflows/stable.yml
PragTob: Having those comments in the code may be nice for our future selves :sweat_smile: I might just add them later though.

spec/return_codes_spec.rb
eregon: I filed https://github.com/oracle/truffleruby/issues/3535 and I'll add a reference to that here then.

eregon: Done.

.github/workflows/stable.yml
eregon: Yeah, fair enough, I'll add a comment here

eregon: Done, should be all good and ready to merge now, thanks for the review!

PragTob:
Heyo, looks good and I fixed JRuby on main so should be good to merge this now. Sorry for all the silence, I wasn't around here.

.github/workflows/stable.yml
PragTob: Thanks for all your working and fixing this up, sorry I wasn't around

PR #1107, opened 2024-08-30
Bring Ruby 2.7 support back

amatsuda:
This PR brings Ruby 2.7 back to the CI matrix by monkey-patching Capybara to correctly switch three `all` methods defined by Capybara, RSpec, and Aruba.

See https://github.com/teamcapybara/capybara/pull/2771 for the details of the Capybara patch.

### DIFF
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -24,6 +24,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - run: rm Gemfile.lock
+
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby-version }}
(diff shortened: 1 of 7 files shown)


amatsuda:
@tycooon Thank you for the question.

> Ruby 2.7 reached EOL on 2023-03-31, why keep supporting it?

IMO testing frameworks and testing tools has to offer long enough support term for real users working on real applications.

It is true that the Ruby team currently cannot support each release no longer than three years, due to lack of maintainer resource, but I know that there are so many real world Ruby applications still running on Ruby 2.x. In fact, I as a Rails consultant used to work on a Ruby 2.4 app until a few months ago, and I'm currently still working on another Ruby 2.7 app.
And for such users, being able to run same set of tests on same set of testing tools is a lifeline for upgrading.

That's the motivation that I want to provide longer support, at least than Ruby/Rails EOL on testing tools (e.g. https://github.com/test-unit/test-unit-rails/pull/32

Issue #1115, opened 2024-11-02
Let's report actionable minimum numbers for Branch Coverage


corsonknowles:
Currently, simplecov is rounding up when delivering branch coverage results. If we then set minimum branch coverage to the reported amount, the next spec run will fail. 

Example:

```bash
Line Coverage: 100.0% (3486 / 3486)
Branch Coverage: 98.2% (763 / 777)
Running RuboCop...
Inspecting 281 files
```

763 / 777 = 0.98198198198

So the ideal number for the report would be 98.19, i.e. the floor of the 2nd decimal place.

Running specs with 98.19 will pass, as expected.

```shell
gem list simplecov

*** LOCAL GEMS ***

simplecov (0.22.0)
simplecov-html (0.13.1, 0.12.3)
simplecov_json_formatter (0.1.4)
```

corsonknowles:
I think we can accomplish this by changing:
https://github.com/simplecov-ruby/simplecov/blob/b6c2d4208a5fa395ce09d7e1d3b074f680ee29b0/lib/simplecov/formatter/simple_formatter.rb#L17C75-L17C80
`output << "#{file.filename} (coverage: #{file.covered_percent.round(2)}%)\n"`

to use `.floor(2)` rather than `round(2)`

corsonknowles:
Digging around (haha, yes pun intended) a little bit, it looks like almost all other occurrences of this issue were handled here:

* https://github.com/simplecov-ruby/simplecov/pull/773/

corsonknowles:
Hi Tobias! 
Can we get your help with this?
@PragTob

corsonknowles:
Here's a potential fix: 
https://github.com/simplecov-ruby/simplecov/pull/1116
(regardless of whether it fixes this specific issue, it moves us closer to consistency and having actionable numbers to set min coverage with)

PR #1116, opened 2024-11-02
Follow the repo convention to use `floor` instead of `round` to provide actionable coverage metrics

corsonknowles:
Use `floor(2)` rather than `round(2)` as we do elsewhere across `Simplcov`

Follow up and relates to:
* https://github.com/simplecov-ruby/simplecov/pull/773/

Relates to and possibly helps with:
* https://github.com/simplecov-ruby/simplecov/issues/1115


amatsuda:
@corsonknowles Good catch! Thank you!

PR #1130, opened 2025-06-08
Merge in simplecov_json_formatter.

simi:
Hello!

I was working on [configuring CI](https://github.com/codeclimate-community/simplecov_json_formatter/pull/11) for `simplecov_json_formatter` and I have realized, it seems probably better to merge it inside the main gem, since it is useless without `SimpleCov` at all (even having no dependency, but it is referencing `SimpleCov` in each file). And on the other side, there is no way to install `SimpleCov` without `simplecov_json_formatter`, since it is direct dependency.

Happy to get proven wrong, but I don't see any reason to not merge it together, since `simplecov_json_formatter` as separate gem felt a little out of radar even it is one of the top gems downloaded per month on RubyGems.org (currently with ~7M monthly downloads).

I have ported JSON formatter 1:1, let the rubocop fix the style and added `verbose: true` additional argument to `format` method to keep original behaviour of putting message to stdout using `puts`, but also to be able to keep it less verbose during testing.

_This gem was identified by [RubyElders](https://github.com/rubyelders) as one of the crucial for whole RubyGems.org ecosystem and it is crucial to get well maintained dependencies. I'm preparing blog.post with more details on this effort._


simi:
@amatsuda WDYT?

PR #1164, opened 2026-04-04
Add richer metrics to JSON formatter

sferik:
Enhance the JSON coverage output with per-file coverage percentages, a total section with aggregate statistics, full group stats, and method coverage support.

Per-file output now includes covered_percent (always), and when enabled: branches_covered_percent and methods array with `methods_covered_percent`. The total and groups sections report full statistics (covered, missed, total, percent, strength) for each enabled coverage type (line, branch, method).

Breaking change: group stats change from `{ covered_percent: 80.0 }` to the full stats shape with the key renamed to `percent`.

Based on the ideas in https://github.com/codeclimate-community/simplecov_json_formatter/pull/12.

PR #1165, opened 2026-04-05
Simplify formatter architecture

sferik:
This PR builds on https://github.com/simplecov-ruby/simplecov/pull/1164 to separate data from presentation in the coverage report by making the HTML formatter a client-side JavaScript (TypeScript) app that reads structured JSON data, rather than a server-side Ruby ERB renderer.

Both formatters share a single entry point for data serialization: `JSONFormatter.build_hash(result)`, which converts a `SimpleCov::Result` into a Ruby hash with no side effects. The JSON formatter calls `build_hash` and writes the result to `coverage.json`. The HTML formatter calls the same method, wraps the result as `window.SIMPLECOV_DATA = {...};` in a JavaScript file, and copies pre-compiled static assets (`index.html`, `application.js`, `application.css`, and favicons) into the output directory. There is no ERB and no Ruby-side rendering.

The JavaScript app reads `window.SIMPLECOV_DATA` on page load and renders everything client-side. Source files are rendered on demand when the user clicks into them.

The ERB templates, Ruby view and coverage helpers, `ResultExporter`, and `SourceFileFormatter` have been deleted, removing about 2,700 lines of code. In their place, a static `index.html` and client-side rendering logic in TypeScript add about 1,500 lines, for a net reduction of roughly 1,200 lines.

The JSON output has been enriched with source code, coverage counts, branch display metadata, group file membership, and result metadata. The formatter class hierarchy has been flattened from 5 classes to 3: `HTMLFormatter`, `JSONFormatter`, and `ResultHashFormatter`.

This gives a clean separation of data and presentation. The HTML report can be regenerated from a saved `coverage.json` without re-running tests via `HTMLFormatter#format_from_json`. The `coverage.json` format becomes a self-contained data API that other tools and third-party formatters can consume directly. The static HTML, JS, and CSS assets are pre-compiled once via `rake assets:compile` and ship with the gem. No runtime template rendering is needed.


spec/fixtures/json/sample_with_branch.json
```diff
@@ -1,6 +1,28 @@
 {
   "meta": {
-    "simplecov_version": "0.22.0"
+    "simplecov_version": "0.22.0",
+    "command_name": "STUB_COMMAND_NAME",
+    "project_name": "STUB_PROJECT_NAME",
+    "timestamp": "2024-01-01T00:00:00+00:00",
+    "root": "/STUB_WORKING_DIRECTORY",
```
brynary, line 7: Since the `root` is being included here as `"/STUB_WORKING_DIRECTORY"`, would it make sense for the coverage paths (e.g. on line 28: `"/STUB_WORKING_DIRECTORY/spec/fixtures/json/sample.rb"` to omit the root and be root-relative?

lib/simplecov/formatter/json_formatter.rb
```diff
@@ -1,36 +1,29 @@
...
-        result_hash = format_result(result)
-
-        export_formatted_result(result_hash)
+      def self.build_hash(result)
+        ResultHashFormatter.new(result).format
+      end
 
+      def format(result)
+        json = JSON.pretty_generate(self.class.build_hash(result))
+        File.write(File.join(SimpleCov.coverage_path, FILENAME), json)
```
brynary: One mistake we see all the time in real life if someone running multiple test processes and not realizing they are overwriting the same coverage datafiles, leaving them with an incomplete set of data. (e.g. when parallelizing tests).

However, when not parallelizing tests, re-running the tests with simplecov enabled is normal and expected behavior. For the parallelization case, a loud warning would be appropriate, but for a non-parallelized case that would be unnecessary and annoying.

I don't have any good recommendations here, just flagging it in case someone else has an idea on if there's a way to cleanly differentiate between "overwrite expected" versus "overwrite == data loss"

spec/fixtures/json/sample_with_branch.json
sferik: Resolved by https://github.com/simplecov-ruby/simplecov/pull/1165/changes/43a5dc548c1cef27736f9b4e3f825f283843d7d3.

lib/simplecov/formatter/json_formatter.rb
sferik: Thanks for flagging this. I took a swing at it in https://github.com/simplecov-ruby/simplecov/pull/1165/changes/43a5dc548c1cef27736f9b4e3f825f283843d7d3.

I use `PROCESS_START_TIME = Time.now` when `JSONFormatter` loads (≈ when this Ruby process required SimpleCov). Before writing `coverage.json`, read the existing file’s `meta.timestamp`. If it’s after `PROCESS_START_TIME`, another process wrote the file while we were running, so warn and point at `SimpleCov::ResultMerger`.

This cleanly differentiates the two cases without any new state:

* Sequential re-run (expected overwrite): the prior file’s timestamp predates the new process’s load.
* Concurrent run (data loss): workers share a similar start time but finish at different times. The later writer sees the earlier writer’s timestamp post-dating its own start.

The write still happens. The warning is advisory—not a hard stop—so non-parallel workflows aren’t blocked. 

If two processes finish within the same floor-to-the-second timestamp, the later writer’s check is `existing_ts > start_time`, so a same-second finish won’t trigger. And if system clocks skew backward between processes, a true concurrent write could look sequential. Neither seemed worth further complexity.

brynary: This seems like a great idea! I agree on advisory and effectively considering this a "best effort" warning. Hopefully someone sees it one time, has an "oh crap" moment and fixes their config.

The only question I would have is -- what do you think about extending the time tracking from seconds to milliseconds precision?

brynary:
I had left two comments but with the new GH UI I can no longer find them:

1. I'm unsure if the `shortenFilename` function may need to account for backslashes on Windows. (Not my jam, just caught my eye.)

2. I would strongly recommend that the relative paths _not_ be prefixed with a forward slash `/`. Consuming tools will likely interpret a forward-slash prefixed path as absolute and do the wrong thing.

sferik:
@brynary thanks for the second round for feedback.

`SourceFile#project_filename` now strips the leading separator so it returns a truly relative path (`lib/foo.rb` instead of `/lib/foo.rb`). That flows through to the `coverage.json` keys, the group file lists, the `errors.minimum_coverage_by_file.*` keys, and the CLI "coverage by file … is below … in <file>" message.

Since paths are now genuinely relative, I removed the `shortenFilename` helper in the frontend entirely. This sidesteps the Windows backslash concern you flagged. The regex in `project_filename` strips `/` or `\`, so the frontend just uses the JSON values as-is.

Users with their own anchored `RegexFilter`s like `%r{^/lib/}` will need to rewrite them as `%r{\Alib/}`. I called this out in the CHANGELOG. This could break a lot of apps, but it’s an easy fix and I agree that it’s more correct and now (ahead of the 1.0 release) is the time to make that change.

lib/simplecov/formatter/json_formatter.rb
sferik: Good call: https://github.com/simplecov-ruby/simplecov/pull/1165/commits/e1f048664d97f55ec31e0e2acad7a1e8aeef0ae8/

features/support/env.rb
```diff
@@ -23,8 +23,23 @@ def extended(base)
 
 # Rack app for Capybara which returns the latest coverage report from Aruba temp project dir
 coverage_dir = File.expand_path("../../tmp/aruba/project/coverage/", __dir__)
+
+# Prevent the browser from caching coverage_data.js between scenario visits
```
PragTob, line 27: clever but necessary 😅  Thanks for the why comment!

html_frontend/src/app.ts
```diff
@@ -39,15 +116,119 @@ function on(
   }
 }
 
+function escapeHTML(str: string): string {
+  const div = document.createElement('div');
+  div.appendChild(document.createTextNode(str));
+  return div.innerHTML;
+}
+
+function md5Hex(str: string): string {
```
PragTob: This is a classic Tobi comment but there's so much md5 code here, I'd prefer to put it into its own file - it distracts from the main purpose imo

html_frontend/src/app.ts
```diff
@@ -39,15 +116,119 @@ function on(
...
+  // Simple string hash that matches Ruby's Digest::MD5.hexdigest output format.
```
PragTob: A small extension of the why comment as to why we need that, would be neat imo (I don't know right now)

html_frontend/src/app.ts
```diff
@@ -68,31 +268,354 @@ function fmtNum(n: number): string {
...
+// --- Rendering: Coverage summary (per source file) ------------
+
+function renderTypeSummary(type: string, label: string, covered: number, total: number, enabled: boolean, opts: { suffix?: string; missedClass?: string; toggle?: boolean } = {}): string {
+  if (!enabled) {
+    return `<div class="t-${type}-summary">\n    ${label}: <span class="coverage-disabled">disabled</span>\n  </div>`;
   }
-  if (numEl) numEl.textContent = fmtNum(covered) + '/';
-  if (denEl) denEl.textContent = fmtNum(total);
+  const missed = total - covered;
+  const pct = total > 0 ? (covered * 100.0 / total) : 100.0;
```
PragTob, line 243: might be worth including those in the output of the JSON formatter imo (JSON a bit bigger, but less derivative work necessary for formatters and services, plus single source of truth for a fix of the metrics)

html_frontend/src/app.ts
```diff
@@ -68,31 +268,354 @@ function fmtNum(n: number): string {
...
+  coveredLines: number, totalLines: number,
+  coveredBranches: number, totalBranches: number,
+  coveredMethods: number, totalMethods: number,
+  branchCoverage: boolean, methodCoverage: boolean,
+  showMethodToggle: boolean
+): string {
+  return '<div class="summary-stats">' +
+    renderTypeSummary('line', 'Line coverage', coveredLines, totalLines, true, { suffix: 'relevant lines covered' }) +
+    renderTypeSummary('branch', 'Branch coverage', coveredBranches, totalBranches, branchCoverage, { missedClass: 'missed-branch-text' }) +
+    renderTypeSummary('method', 'Method coverage', coveredMethods, totalMethods, methodCoverage, { missedClass: 'missed-method-text-color', toggle: showMethodToggle }) +
```
PragTob, line 272: I like the breakdown of them all to use the same helper, great job 🚀

lib/simplecov/formatter/json_formatter.rb
```diff
@@ -1,34 +1,60 @@
...
   module Formatter
     class JSONFormatter
+      FILENAME = "coverage.json"
+
+      # Captured when the class loads, which is close enough to "when this
+      # Ruby process started tracking coverage". Used to detect whether an
+      # existing coverage.json was written by a sibling process running
+      # concurrently (see #warn_if_concurrent_overwrite).
+      PROCESS_START_TIME = Time.now
+      private_constant :PROCESS_START_TIME
```
PragTob: Can't we set it during `SimpleCov.start` and keep it as part of the state?

CHANGELOG.md
```diff
@@ -5,9 +5,12 @@ Unreleased
...
 * JSON formatter: `simplecov_json_formatter` gem is now built in. `require "simplecov_json_formatter"` continues to work via a shim.
 * `StringFilter` now matches at path-segment boundaries. `"lib"` matches `/lib/` but no longer matches `/library/`. Use a `Regexp` filter for substring matching.
+* `SourceFile#project_filename` now returns a truly relative path with no leading separator (e.g. `lib/foo.rb` instead of `/lib/foo.rb`). This also removes the leading `/` from file path keys in `coverage.json` and from the filename in `minimum_coverage_by_file` error messages. Anchored `RegexFilter`s that relied on a leading `/` (e.g. `%r{^/lib/}`) should be rewritten (e.g. `%r{\Alib/}`).
 * Removed `docile` gem dependency. The `SimpleCov.configure` DSL block is now evaluated via `instance_exec` with instance variable proxying.
+* Removed automatic activation of `JSONFormatter` when the `CC_TEST_REPORTER_ID` environment variable is set. The default `HTMLFormatter` now always writes `coverage.json`, so the env-var special case is no longer needed.
```
PragTob: the JSON formatter is writing the `coverage.json`, right?

html_frontend/src/app.ts
PragTob: although... isn't some of that in the formatter alreeady?

PragTob (commented):
First, I want to thank you for your work you put into this. 💚 

That said, I'm still not a fan of the approach of letting the web frontend read the JSON again:

This moves a lot of core logic of simplecov to TypeScript, which for a Ruby project makes it less friendly to contribute to and fix (while many Ruby folks know JS/TS, not all do and often not as well).

I see the point of rendering on just data/a known data exchange format. But, since we already have the data loaded in Ruby with helpers et. al. it's just as fine to render it from Ruby in my opinion - there are no extra steps necessary (we're not talking to a REST API or something, but dealing with data we just gathered in our own process).

Rendering this out on the SimpleCov side _may_ be faster than the old HTML renderer. That said, I liked that it was done then and not much rendering had to be done on page load. Call me old-fashioned but for a mostly static report I appreciate most of it being rendered once.

I was gonna say "but I can't argue with the reduction in loc", but I think the following claim is misleading:

> The ERB templates, Ruby view and coverage helpers, ResultExporter, and SourceFileFormatter have been deleted, removing about 2,700 lines of code. In their place, a static index.html and client-side rendering logic in TypeScript add about 1,500 lines, for a net reduction of roughly 1,200 lines.

It makes it seem like the new approach is much more efficient in terms of line of code, that is not true.
The reduction comes from deleting the tests, not the source files. A net of 1700+ of the deletions is the test files for the removed ruby code, the net of the removed ruby code being ~400 loc (less than the TS code added, although granted a lot of that is also the md5 function so I suppose we can call it equal).

```
FOLDER                         ADDS       DELS
html_frontend                  678        57
Rakefile                       4          1
CHANGELOG.md                   3          0
spec                           290        81
features                       44         78
lib                            240        649
README.md                      0          3
test                           182        1926
```

The JS code also no equivalent tests, so for an argument on lines deleted/added we could have deleted just the ruby tests and had an equivalent reduction.

I didn't write nor review many of the html formatter tests, but this also means less tests to rely on/more reliance on the feature tests.

I'd prefer to keep the rendering logic on the ruby side, mostly to make it easier for people to debug and contribute - while having unit tests for it is also nice (I know we _could_ also write some for the TS side).

I'll repeat what I said before, that's my opinion but imo I haven't done enough work on SimpleCov recently to push on that.

Other comments are mostly optional, although I'd be interested on the time stamp one. 

Image

CHANGELOG.md
sferik: `HTMLFormatter` emits `coverage.json` by calling `JSONFormatter.build_hash`, producing the same payload as `JSONFormatter`. Clarified in https://github.com/simplecov-ruby/simplecov/pull/1165/changes/f428f5a1bd063bb3ca320f4dac3d79faa06d03c3.

lib/simplecov/formatter/json_formatter.rb
sferik: Great idea! Implemented in https://github.com/simplecov-ruby/simplecov/pull/1165/changes/91ade9d18d2291cc4a0157c886553553c6e865f3.

html_frontend/src/app.ts
sferik: Resolved in https://github.com/simplecov-ruby/simplecov/pull/1165/changes/fffeaf5f1ce5c28ad78ea38ac0b8bb147f58dde9.

html_frontend/src/app.ts
sferik: “Why” comment added here in https://github.com/simplecov-ruby/simplecov/pull/1165/changes/7363ba3d1635fb4d053ccb795935e9e264a5c80b.

html_frontend/src/app.ts
sferik: Originally I had all this MD5 code here to maintain backward compatibility with `Digest::MD5.hexdigest`, but I don’t think may people bookmark SimpleCov deep links and 1.0 is the time to break things anyway, so I’ve replaced MD5 with truncated SHA-1 computed in the browser via Web Crypto. I also moved it into a separate file (`hash.ts`), which is only 10 lines of code. See https://github.com/simplecov-ruby/simplecov/pull/1165/changes/71514a940b3eeb537cb50843592d9a569143389c.

sferik:
@PragTob Thanks for the thoughtful review. Your inline comments were very helpful and I’ve addressed all of them with code. Your broader points were also good, so let me try to address those…

On contributor-friendliness: just because SimpleCov is a tool for the Ruby community, I don’t think 100% of the code needs to be in Ruby/ERB. In fact, this has never been the case; there has always been some amount of JavaScript in SimpleCov. The difference is that now the Ruby/JS boundary corresponds to a real architectural seam rather than arbitrary file boundaries. Anyone fixing how coverage is measured or attributed still works exclusively in Ruby. And thanks to your feedback, https://github.com/simplecov-ruby/simplecov/pull/1165/changes/fffeaf5f1ce5c28ad78ea38ac0b8bb147f58dde9 moves the last bits of frontend recomputation back into `ResultHashFormatter`, so the formatter is the single source of truth for every derived metric.

On the render-once-vs-on-page-load preference: you’re right that the TS does run at page load, and I don't want to pretend that’s not a trade. But I think the “render once” intuition is sized for the old ERB renderer, where rendering meant computing all the metrics. The TypeScript code isn’t doing that. The JSON formatter already has, so the TS just walks the data and builds DOM. Even on large reports it’s milliseconds. What we get in exchange is `HTMLFormatter#format_from_json`, which can rebuild the HTML view from any existing `coverage.json`. The same thing benefits any third-party tool: services, custom formatters, dashboards now get the same payload the HTML view does, instead of having to re-derive the numbers themselves. Previously, the output of `JSONFormatter` and the HTML report had drifted in shape, this architecture ensures that won’t happen again in the future.

You were also right that the LOC framing in the description was misleading. I shouldn’t have leaned on it. Honestly, the TS is roughly comparable in size to the deleted ERB + view helpers, but the report itself does quite a bit more (sortable/filterable tables, keyboard nav, in-page filename filter, dark mode, on-demand source rendering). The actual win isn't fewer lines, it’s one data path for many consumers.

The test-coverage point is fair too. The data layer is still fully unit-tested in Ruby, and the HTML formatter tests now assert against the parsed JSON payload, but you're right that there’s no direct unit coverage for the TS modules. Cucumber/Capybara is the only thing exercising the rendered DOM. If that’s a gate for you before 1.0, I'm happy to add Vitest specs.

PragTob:
👋

Thanks for all your work and all my tardiness. It's been a couple of very intense weeks.

Also thanks for addressing the comments so diligently and promptly, great improvements 👌 

For the broader topics:

>  I don’t think 100% of the code needs to be in Ruby/ERB. In fact, this has never been the case; there has always been some amount of JavaScript in SimpleCov.

Never meant to say it should be or it was. But we're migrating it here, I don't see a distinct reason why TS would be better than Ruby here and all else being equal I'd default to keeping code in Ruby inside a Ruby project. JS/TS for the parts that benefit from it/are needed for interactivity/usage of the HTML report.

> The same thing benefits any third-party tool: services, custom formatters, dashboards now get the same payload the HTML view does, instead of having to re-derive the numbers themselves. (...)
> The actual win isn't fewer lines, it’s one data path for many consumers.

I'd argue that was the case previously. The API was the Ruby structures simplecov built and you got access to when running a formatter. Any important metric not accessible via a nice helper there should be considered a bug and added to the API of the class. That way everyone had the data they needed easily at their fingertips and could easily do with it whatever (create JSON, XML, POST to an API...).

That some folks refused the offered contract/utilities is not the fault of SimpleCov.

Feels like we're changing a lot (including a JSON formatter in SimpleCov, producing the JSON by default for everyone, moving the HTML formatter from Ruby to TS/JS) to help support one very specific case/ask which doesn't feel right to me. But we've had that discussion.

> The test-coverage point is fair too. (...) If that’s a gate for you before 1.0, I'm happy to add Vitest specs.

I didn't mean to imply that. Imo this is fair/it's not too much code and it doesn't seem vital to be unit tested esp. since we have a good chunk of cucumber tests. I was just very irritated by it being called equal with lines of code when tests were deleted without replacement, hence pointing it out. Are unit tests nice and can help? Sure. Are they vital in the current circumstances? I don't think so.


----------------------------------------------------

We probably won't be able to agree what's the right approach here, and that's fine. As a fan of data over functions I can see the benefit of having an official data exchange format. I don't see it being run as the default and changing the formatter to accommodate that as the best choice. The "source of truth" for me are still the Ruby objects - if you want JSON, configure a JSON formatter et voila you got JSON. That said, it's also very hard for me to shake the feeling that we're doing this majorly for the benefit of a company that had previously made bad choices and then chose to treat this project poorly and force poor choices on it in turn. As such, my vision is probably also somewhat clouded. 

You seem convinced that this is the best way forward, want to push and support it and are doing to work to make it happen, for which I'm thankful. So, make it happen :)

sferik:
I just got off a call with @PragTob and he agreed that we could try this approach. 😊

Issue #56, opened 2011-07-08
Only String accepted for source (ArgumentError)


ryanjm:
I've run into this twice now, and I'm sorry, but I don't know what's tripping it up.

I'm using guard to run my tests / simplecov.

If it runs with this code:

```
def following_all
  ...
  all.map! do |a|
    # Return array
    if a.class == MyClass
      ... do something
    else
      ... do something else
    end
  end
  all.flatten!
end
```

It will be fine. But if I take out that the line with the comment (the comment doesn't matter, the extra line does) then it will crash.

```
/Users/ryanjm/.rvm/gems/ruby-1.9.2-head/gems/simplecov-0.4.2/lib/simplecov/source_file.rb:25:in `initialize': Only String accepted for source (ArgumentError)
    from /Users/ryanjm/.rvm/gems/ruby-1.9.2-head/gems/simplecov-0.4.2/lib/simplecov/source_file.rb:66:in `new'
    from /Users/ryanjm/.rvm/gems/ruby-1.9.2-head/gems/simplecov-0.4.2/lib/simplecov/source_file.rb:66:in `block in lines'
    from /Users/ryanjm/.rvm/gems/ruby-1.9.2-head/gems/simplecov-0.4.2/lib/simplecov/source_file.rb:65:in `each'
    from /Users/ryanjm/.rvm/gems/ruby-1.9.2-head/gems/simplecov-0.4.2/lib/simplecov/source_file.rb:65:in `each_with_index'
    from /Users/ryanjm/.rvm/gems/ruby-1.9.2-head/gems/simplecov-0.4.2/lib/simplecov/source_file.rb:65:in `lines'
    from /Users/ryanjm/.rvm/gems/ruby-1.9.2-head/gems/simplecov-0.4.2/lib/simplecov/source_file.rb:91:in `missed_lines'
    from /Users/ryanjm/.rvm/gems/ruby-1.9.2-head/gems/simplecov-html-0.4.5/lib/simplecov-html.rb:80:in `block in lines_missed'
    from /Users/ryanjm/.rvm/gems/ruby-1.9.2-head/gems/simplecov-html-0.4.5/lib/simplecov-html.rb:80:in `map'
    from /Users/ryanjm/.rvm/gems/ruby-1.9.2-head/gems/simplecov-html-0.4.5/lib/simplecov-html.rb:80:in `lines_missed'
[...]

PR #1172, opened 2026-04-17
Drop warning and ruby 1.9 support.

### LINKED ISSUE #1057: option to disable line warnings with `enable_coverage_for_eval` and rails views
Couldn't find an issue that was directly related to this, please forgive me if I've overlooked something.

Love `enable_coverage_for_eval` since that allows getting coverage metrics for rails views, which was something that until Ruby 3.2, you had to do without.  However, turning it on results in a _lot_ of warnings, relative to this method in lib/simplecov/source_file.rb:

```ruby
# lib/simplecov/source_file.rb:250-253
    # Warning to identify condition from Issue #56
    def coverage_exceeding_source_warn
      warn "Warning: coverage data provided by Coverage [#{coverage_data['lines'].size}] exceeds number of lines in #{filename} [#{src.size}]"
    end
```

Reviewing Issue#56, it wasn't clear why this warning occurs, other than we know that we're getting more lines reported for a source file than the code in the file actually has.  I have not been able to determine why this is, but suspect it might be due to erb inserting lines in the file.

Below is a copy of the patch I created to allow turning these warnings off with an ENV var.  Granted this is a crude fix, but it helps to clean up our test output.

My request is getting the below patch or something like it included in simplecov.  Ultimately, these warnings are not helpful in their current form for view specs, since it's not clear why this is an issue or what can be done to resolve it.

```shell
$ ruby -e "puts RUBY_DESCRIPTION"
ruby 3.2.0 (2022-12-25 revision a528908271) [x86_64-linux]
```
Rails 7.0.4.3
RSpec 3.12.0
SimpleCov 0.22.0

tests are run with `bin/rspec spec` manually or via guard.

```ruby
# spec/spec_helper.rb
# frozen_string_literal: true

unless RSpec.configuration.files_to_run.one?
  require 'simplecov'
  require_relative 'support/simplecov_warnings_patch'
  SimpleCov.start 'rails' do
    enable_coverage :branch
    enable_coverage_for_eval
    # minimum_coverage line: 85, branch: 70
    # maximum_coverage_drop 5
  end
end

...
<more spec_helper>

# spec/support/simplecov_warnings_patch.rb
# frozen_string_literal: true

module SimpleCov
  class SourceFile
    # this is needed for SimpleCov Issue#56
    # it also shows up for views when we `enable_coverage_for_eval`
    # patching here so that we can turn this off via ENV VAR
    def coverage_exceeding_source_warn
      return unless ENV['SHOW_EXCESS_LINE_WARNING']

      message = "Warning: coverage data provided by Coverage [#{coverage_data['lines'].size}] "
      message += "exceeds number of lines in #{filename} [#{src.size}]"
      warn message
    end
  end
end
```

This is an example view spec for devise (v4.8.1) that reproduces the issue (remove the `unless` wrapper on SimpleCov config to get results from specs for a single file):
```ruby
# spec/views/devise/registrations/new.html.erb_spec.rb
# frozen_string_literal: true

require 'rails_helper'

RSpec.describe 'devise/registrations/new.html.erb' do
  let(:resource) { build_stubbed :user }
  let(:devise_mapping) { Devise.mappings[:user] }

  before do
    render template: 'devise/registrations/new', \
           locals: { resource:, resource_name: :user, devise_mapping: }
  end

  it 'accepts username' do
    expect(rendered).to have_css '#user_username'
  end

  it 'accepts email' do
    expect(rendered).to have_css '#user_email'
  end
end
```

Again, main focus here is to have an env var or config option added to simplecov that allows silencing the warnings from the `coverage_exceeding_source_warn` method.

coorasse:
From my understanding the warning was introduced in the issue #56 and it's interesting only for ruby 1.9, which is not supported anymore anyway.
So let's get rid of this warning completely and close #1057 as well.


sferik (changes requested):
I would approve this after a few small changes:
1. I’m not sure why you added `spec/fixtures/issue_1057.erb`, but it doesn’t seem to be referenced anywhere, so please either use it or remove it.
2. The `capture_stderr` is now unused and can be removed.
3. Add a positive regression test: `expect { subject.lines }.not_to output.to_stderr` so silence is pinned.

sferik:
@coorasse Are you planning to make these changes so this PR can be merged or would you like me to take it over?

coorasse:
On it right now 👍 thanks

coorasse:
Adapted. Sorry for the `issue_1057.erb`. I used it in a previous iteration and then removed it.

sferik (approved):

sferik:
Merged! Thanks so much.

PR #1190, opened 2026-05-26
Document explicit receiver in .simplecov file

bquorning:
In the next version of simplecov, all method calls in .simplecov must have an explicit `SimpleCov` receiver.

I tested out the recent changes in https://github.com/rubocop/rubocop-rspec/pull/2177 and found that removing the deprecated `SimpleCov.start` and _not_ adding an explicit receiver ended up in warnings about various methods not being defined on `main`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,8 +319,8 @@ starts it:
 ```ruby
 # .simplecov — configuration only
 SimpleCov.profiles.load 'rails'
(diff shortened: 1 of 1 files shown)


sferik (approved):
Thanks for testing this out and catching and fix this documentation bug!

PR #1193, opened 2026-05-26
README: Update to preferable coverage pattern for Rails

runephilosof-abtion:
Preferred Rails configuration, as noted in the [Rails profile](https://github.com/simplecov-ruby/simplecov/blob/main/lib/simplecov/profiles/rails.rb#L22)


sferik (approved):
Nice catch. Thanks!

PR #1216, opened 2026-06-26
Add merge finalization ownership config

### LINKED ISSUE #1215: Parallel workers with sharded coverage_dir wait before external collate
## Summary

A user reported what looks like a hang when SimpleCov is used with parallel test workers that each write coverage to an isolated coverage directory, followed by an explicit `SimpleCov.collate` cleanup step.

Repro repository: https://github.com/G-Rath/simplecov-wat

Fix branch: https://github.com/kettle-dev/simplecov/tree/fix/finalize-merge-ownership

## Reproduction

In the repro, `spec/spec_helper.rb` enables merging inside parallel workers and writes each worker to its own coverage directory:

```ruby
if ENV.key?("TEST_ENV_NUMBER")
  merging true
  coverage_dir "coverage/turbo_tests/#{ENV["TEST_ENV_NUMBER"]}"
  command_name "rspec-#{ENV["TEST_ENV_NUMBER"]}"
  formatter SimpleCov::Formatter::SimpleFormatter
end
```

The `Rakefile` then owns final report generation with an explicit cleanup collate step:

```ruby
SimpleCov.collate Dir["coverage/turbo_tests/*/.resultset.json"] do
  coverage(:line) { minimum 100 }
  coverage(:branch) { minimum 100 }
end
```

Running plain RSpec completes normally:

```sh
bundle exec rspec
```

Running the parallel path reproduced the issue before the fix:

```sh
bundle exec turbo_tests2 spec
```

## Actual behavior before the fix

The tests themselves completed, but SimpleCov waited for `parallel_wait_timeout` (60 seconds by default) before continuing. The warning showed only one worker result was visible from the worker that SimpleCov selected to finalize the report, even though all workers had written their own shard resultsets.

## Root cause

`merging true` currently means two distinct things:

1. Store this process' coverage resultset so it can be merged later.
2. Let this process own final merge/report processing: wait for siblings, build the merged result, format, enforce thresholds, and write `.last_run.json`.

Those two responsibilities conflict when workers intentionally write isolated resultsets such as:

```text
coverage/turbo_tests/1/.resultset.json
coverage/turbo_tests/2/.resultset.json
coverage/turbo_tests/3/.resultset.json
```

The final worker's `ResultMerger.resultset_path` points only at its own shard. Its polling loop waits for all expected workers to appear in that one file, but sibling workers are writing different files by design. The later `SimpleCov.collate` cleanup step is the true finalizer in this setup, so worker-side finalization should not run.

## Proposed / implemented fix

Add `SimpleCov.finalize_merge` to separate storing resultsets from owning final report finalization.

Semantics:

- `merging true` still means workers store mergeable resultsets.
- `finalize_merge true` means this process owns wait / merge / format / threshold / `.last_run.json` finalization.
- `finalize_merge false` means this process stores its resultset but does not wait, merge, format, enforce thresholds, or write `.last_run.json`.
- `SimpleCov.collate` remains a finalizer and writes `.last_run.json` for the collated result when coverage checks pass.

For compatibility, the branch keeps the default as `true`, but conservatively infers `false` with a configuration warning only when all of these are true:

1. A recognized parallel adapter is active.
2. More than one worker is expected.
3. Merging is enabled.
4. The coverage destination was explicitly changed from the default.
5. Parallel-worker environment is present.

The warning tells users to set either `SimpleCov.finalize_merge false` to keep external collation ownership or `SimpleCov.finalize_merge true` if the selected worker should own the built-in finalization flow.

## Validation

Validated on the branch:

- Focused SimpleCov specs for configuration, worker gating, and collate finalization.
- Full SimpleCov spec suite.
- RuboCop.
- The repro repo with local SimpleCov.

The repro's `bundle exec turbo_tests2 spec` now finishes in about one second, `SimpleCov.collate` generates the final report for `rspec-1, rspec-2, rspec-3`, and the final `coverage/.last_run.json` is written by the collate step rather than by workers.

pboling:
## Summary

Adds `SimpleCov.finalize_merge` so parallel workers can store mergeable resultsets without also owning final report finalization.

Fixes #1215.

## Root cause

`use_merging true` currently covers two responsibilities: storing resultsets and finalizing the merged report. In setups like the [repro](https://github.com/G-Rath/simplecov-wat), each worker writes to an isolated `coverage_dir` and a later `SimpleCov.collate` cleanup step owns the final report. The selected final worker was polling only its own shard `.resultset.json`, so it waited for sibling workers that were intentionally writing different files.

## Changes

- Add `SimpleCov.finalize_merge` with explicit true/false support.
- Keep the default true, but infer false with a configuration warning for recognized multi-worker parallel runs with merging enabled and an explicit custom coverage destination.
- Gate worker wait/merge/format/threshold/`.last_run.json` finalization when `finalize_merge false`.
- Preserve `SimpleCov.collate` as a finalizer that writes `.last_run.json` for the collated result.
- Document merge finalization ownership using the v1 `coverage` configuration API and add changelog coverage.

## Validation

- Running RuboCop

```text
Inspecting 196 files
....................................................................................................................................................................................................

196 files inspected, no offenses detected
```

- `rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb`

```text
Randomized with seed 20626
......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................*....................*.................................................................................................................................................

Non-app warnings written to tmp/warnings.txt

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) SimpleCov::StaticCoverageExtractor.real_source_positions when Prism is not available returns nil so the eval_generated filter is a no-op
     # Prism is available; the no-Prism path can't be exercised on this Ruby
     # ./spec/static_coverage_extractor_spec.rb:190

  2) SimpleCov::StaticCoverageExtractor.call when Prism is not available returns nil so callers fall back to empty hashes
     # Prism is available; the no-Prism path can't be exercised on this Ruby
     # ./spec/static_coverage_extractor_spec.rb:18

Finished in 11.75 seconds (files took 0.54538 seconds to load)
1149 examples, 0 failures, 2 pending

Randomized with seed 20626
```

- Repro: `bundle exec turbo_tests2 spec` in https://github.com/G-Rath/simplecov-wat finishes in about 1 second and collates the final report successfully.


lib/simplecov/configuration/merging.rb
```diff
@@ -83,5 +115,55 @@ def parallel_wait_timeout(seconds = nil)
       @parallel_wait_timeout = seconds if seconds.is_a?(Integer)
       @parallel_wait_timeout ||= 60
     end
+
+  private
+
+    def inferred_finalize_merge?
+      return true unless merging_enabled_for_inference?
```
sferik: Why not:

```suggestion
      return true unless merging
```

Then I believe you could delete the `merging_enabled_for_inference?` method below, which does the exact same thing. I also think it’s better not to mutate an instance variable in a predicate method.

sferik (changes requested):
The explicit `finalize_merge` API is clean and I’m 👍 on that part. I have some reservations about the auto-inference, but I’m willing to put it out in an RC and see what comes back.

My only nit is that the call to `merging_enabled_for_inference?` could be replaced with `merging`, at which point I believe `merging_enabled_for_inference?` could be deleted since it’s no longer called anywhere.

lib/simplecov/configuration/merging.rb
pboling: Agree. Fixed in [a42450e](https://github.com/simplecov-ruby/simplecov/pull/1216/commits/a42450e06c71de1b3cbd750928a61aedbcbba509)

PR #1223, opened 2026-07-11
Reproduce turbo_tests2 external collation output

pboling:
This is a regression-style test for parallel test output as reported here:
https://github.com/galtzo-floss/turbo_tests2/issues/15

This is not a current issue, it has already been fixed in v1 release candidates.

Related: https://github.com/simplecov-ruby/simplecov/issues/1155


sferik (approved):
Looks good. Thanks!

PR #1225, opened 2026-07-13
Explicitly write formatter's status message to STDERR

viralpraxis:
Using `Kernel.warn` seems it be an overkill here -- some setups track warnings during running test suite; these "warnings" are unaddressable and noisy.


sferik (approved):
This looks good. I’m going to add a regression spec after merging it to ensure the change isn’t reverted in the future.

sferik:
Shipped in [SimpleCov v1.0.1](https://github.com/simplecov-ruby/simplecov/releases/tag/v1.0.1).

PR #1239, opened 2026-07-22
fix: memoise key parsing in RubyDataParser to fix combiner performance regression

oleksii-leonov:
Measured on a real 66-worker CI merge (~21k files, ~46k branches, branch coverage enabled): the collate step (which merged 66 reports) regressed from ~29s under `0.22` to ~145s (5x slower) under `1.0.0`.

A wall-clock stackprof attributes main time to `Ripper#parse` under `RubyDataParser.parse_array_string`.

Resultset JSON stores keys as stringified Ruby literals (`"[:if, 0, 14, 4, 14, 30]"`), and `RubyDataParser` walks them back into arrays with a Ripper parse. That parse sat on a hot path: `BranchesCombiner` and `MethodsCombiner` derive a merge identity from every key of *both* sides on *every* pairwise merge. Folding N resultsets therefore parsed each key string N-1 times.

This PR adds memoisation to `RubyDataParser.call` to avoid parsing strings that have already been parsed.

Notes:

- The set of unique keys is bounded by the project's branch and method count, so the cache stays reasonably small and needs no eviction.
- Cached arrays are frozen and shared. Every caller (both combiners, `BranchBuilder`, `MethodBuilder`, `SourceFile::Method`) destructures the parsed array without mutating it; freezing enforces that any future mutation surfaces as an error instead of silent cross-caller corruption.
- Array inputs still pass through untouched and unfrozen (in-process results never hit the cache, exactly as before).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Unreleased
 ## Bugfixes
 * `SimpleCov.formatter` and `SimpleCov.formatters` now accept formatter instances in addition to formatter classes, so constructor options can actually be passed — most notably `SimpleCov::Formatter::HTMLFormatter.new(silent: true)` to suppress the "Coverage report generated" status line. Previously SimpleCov unconditionally called `.new` on whatever was configured, so passing an instance crashed with `NoMethodError` at report time. See #1240.
 
+## Performance
+* Fix 5x performance regression on report combining (introduced in `1.0.0` as a result of using `Ripper#parse` in a hot path) by adding parsed key memoisation to `RubyDataParser.call`.
+
 1.0.2 (2026-07-18)
 ==================
 
--- a/lib/simplecov/source_file/ruby_data_parser.rb
+++ b/lib/simplecov/source_file/ruby_data_parser.rb
@@ -15,10 +15,28 @@ module SimpleCov
 
       # Tests use the real data structures (except for integration tests)
       # so no need to put them through here.
+      #
+      # String parses are memoized: `Combine::BranchesCombiner` and
+      # `Combine::MethodsCombiner` derive a merge identity from every key of
+      # both sides on every pairwise merge, so collating N resultsets parses
+      # each key string N-1 times — and Ripper dominates the wall time of a
+      # large collate.
+      # Key strings repeat across folds and within report building, while the
+      # set of unique keys is bounded by the project's branch and method
+      # count, so a permanent cache stays small. Cached arrays are frozen
+      # element-wise (String class names included, not just the tuple):
+      # every caller destructures without mutating, and sharing one array
+      # across callers must stay that way. The cache key needs no such
+      # care — `Hash#[]=` dups and freezes String keys on its own.
       def call(structure)
         return structure if structure.is_a?(Array)
 
-        parse_array_string(structure.to_s)
+        string = structure.to_s
+        parse_cache[string] ||= parse_array_string(string).each(&:freeze).freeze
+      end
+
+      def parse_cache
+        @parse_cache ||= {} #: Hash[String, untyped]
       end
 
       # Parse a string like '[:if, 0, 3, 4, 3, 21]' or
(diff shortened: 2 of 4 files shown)


sferik:
@oleksii-leonov Have you tested how this patch impacts the performance on your 66-worker CI merge?

oleksii-leonov:
@sferik 

Yes, we are using this branch in our CI.

Merge time dropped from ~150 seconds to ~35 seconds (roughly the same as it was for 0.22).

oleksii-leonov:
I will check Copilot comment and rebase this PR (there is a conflict in CHANGELOG.md).

sferik (approved):
Approved. I will merge as soon as the CHANGELOG conflict is resolved.

oleksii-leonov:
@sferik , the conflict was resolved. Feel free to merge.

I have a second proposal with some more refactoring (moving from pairwise merge into one-pass merge using a single accumulator instance). In our CI, it cut the time almost in half again (~35 seconds to ~19 seconds; time left is mainly JSON parsing).

sferik:
@oleksii-leonov I prefer to "Rebase and merge" and GitHub still says “This branch cannot be rebased due to conflicts”. To fix that, you’ll need to do the following:

```
git fetch upstream # your remote for simplecov-ruby/simplecov
git checkout perf/memoize-coverage-key-parsing
git rebase upstream/main
```

The rebase will stop on your first commit with a conflict in `CHANGELOG.md`. Please keep both sides. Then:

```
git add CHANGELOG.md
git rebase --continue
git push --force
```

When do you expect your second PR will be ready? I was planning to push v1.0.3 tomorrow, but I’m happy to wait for your second patch to land.

Thanks so much for making these performance improvements!

oleksii-leonov:
Oh, got it. I will make a clean rebase and squash it into 1 commit.

oleksii-leonov:
@sferik , rebase and squash done.

> When do you expect your second PR will be ready? I was planning to push v1.0.3 tomorrow, but I’m happy to wait for your second patch to land.

I would say let's merge/publish this PR first (it's simple and solves the main regression issue).

The second change is more complicated (https://github.com/simplecov-ruby/simplecov/commit/60a36ee0d33ad797821ce2783dc9ef00c7bbac01). Specs are green, and results on our codebase are identical. I am already using it in our CI to battle test. So I would prefer to merge/publish it separately.

sferik:
@oleksii-leonov Thanks for the contribution! This change shipped in [v1.0.3](https://github.com/simplecov-ruby/simplecov/releases/tag/v1.0.3).

I also took a look at 60a36ee0d33ad797821ce2783dc9ef00c7bbac01 and it looks mostly good. I verified the fold-vs-accumulator equivalence. A couple things before you open a PR:

1. I found one parity bug. When the executed-side-is-authoritative rule engages, the fold substitutes NO_SYNTHESIZED (empty hashes), so the merged file always keeps at-least-empty "branches" and "methods" tables. The accumulator leaves them `nil` in both drop paths (`drop_synthesized`, and the `elsif @executed` early return for an incoming simulated result). When no executed result carries a "methods" key (criteria drift between workers), the merged resultset then contains `"methods": null`, which crashes report building in `MethodBuilder` (`fetch("methods", none)` returns the `nil` for a present key, then `nil.filter_map` raises). You should be able to fix this by assigning empty tables in `drop_synthesized` and ensure them before the early return.

2. Please add tests. At minimum the simulated and executed orderings through `FileAccumulator` and ideally a differential spec fuzzing the fold against the accumulator.

Happy to review the PR once these are addressed!

oleksii-leonov:
@sferik , thank you a lot!

Regarding https://github.com/simplecov-ruby/simplecov/commit/60a36ee0d33ad797821ce2783dc9ef00c7bbac01 , I will try to shape it better in the following weeks (currently it's a draft).

sferik:
@oleksii-leonov You may be interested in https://github.com/simplecov-ruby/simplecov/pull/1244, which further improves performance beyond this PR and composes well with 60a36ee0d33ad797821ce2783dc9ef00c7bbac01 (although there may be some minor conflicts, so you’ll want to rebase after I merge it).

PR #1246, opened 2026-07-31
Speed up SimpleCov.collate with optional multi-process merge

danielwestendorf:
This is an AI-Assisted PR

Collating the resultsets from a large CI matrix is slow in a way that scales badly. The collating process reads, parses and folds every shard in sequence, and nearly all the wall clock goes into that fold. This PR let's this process run on more than one core, speeding up overall collation.

Tested on real CI 1 process vs multi-processes (CI node cpu count = 4) for 160 real resultsets on a 231k line rails app saw `collate` reduce from 49sec to ~30sec.

```
┌────────────────────────┬─────────────┬──────┐
│                        │ merge time  │      │
├────────────────────────┼─────────────┼──────┤
│ single process         │ 49s         │      │
├────────────────────────┼─────────────┼──────┤
│ processes: 4           │ 30s         │ −61% │
├────────────────────────┼─────────────┼──────┤
│ processes: 6           │ 28s         │ −77% │
└────────────────────────┴─────────────┴──────┘
```

image
image

### The API change

SimpleCov.collate takes a new processes: argument:

```ruby
SimpleCov.collate Dir["simplecov-resultset-*/.resultset.json"], processes: 6
```

It defaults to 1, which never forks, so existing calls are completely unaffected — same code path as before. Everything else about the signature is unchanged.

Early attempts used `Etc.nprocessors` to automatically tune process count, however, this often reports the number of processes of the host machine, not the vm/container of which it runs, so I opted for manual tuning. Setting ENV var `SIMPLECOV_CONCURRENCY` will also work.

### Failure handling

Every failure path returns nil rather than a partial merge, and the caller redoes the fold in-process. Reporting coverage for a subset of the resultsets would silently understate it, which is a worse outcome than being slow. That covers:

- a runtime that cannot fork — JRuby, TruffleRuby, Windows.
- a worker that died
- a payload that came back truncated

Workers ship their folded pair back over a pipe, deserialized on a thread per worker so every pipe is drained while the workers are still writing — otherwise a payload larger than the pipe buffer blocks its worker mid-write and the parent blocks reaping a worker that can never finish. Children end with exit! so they never fall through to the collating process's at_exit handlers. Each worker folds its slice one file at a time, as the serial path does, so memory scales with the worker count rather than the resultset count.


sferik (changes requested):
Thanks for this PR! Overall, it looks quite good.

I agree with the first three Copilot comments. I’m not sure it’s necessary to coerce `processes` to an Integer (the fourth comment from Copilot), so feel free to skip that one.

Beyond what Copilot caught:

1. The differential test never exercises the parallel association. The "identical, not merely equivalent" claim depends on the combiners being associative: serial computes `((a+b)+c)+d` while the fan-out computes `(a+b)+(c+d)`. Same visiting order, different grouping. `combine_differential_spec.rb` only compares the serial `:fold` against the reference oracle, and the `strategy` parameter in `mismatch_for` is only ever called with `:fold`, which suggests a chunked strategy existed at some point and was dropped. The "same pair however many processes" spec uses simple line-only fixtures, so none of the adversarial `reconcile_synthesized` or drifted-key cases the fuzzer generates ever go through the chunked path. Adding a `:chunked` strategy (chunk, fold each slice, fold the slices) would directly pin the property the whole PR rests on with relatively little code.

2. The `succeeded?(-1)` spec is fragile. `Process.wait2(-1)` doesn't mean "a pid I can't reap", it means "wait for any child". If an unrelated child process is alive when that example runs, the call can block or reap a stranger's child and return `true`. It only passes today because nothing else leaves children around. A guaranteed-nonexistent positive pid (or stubbing `wait2` to raise `Errno::ECHILD`) would make it more deterministic.

3. Naming nit: `merge_fuzzer.rb` and `merge_reference.rb` both say they serve `spec/merge_differential_spec.rb`, but the actual file is `spec/combine_differential_spec.rb`.

The PR description says non-fork runtimes are detected by the `NotImplementedError` the call raises rather than by `respond_to?`, but the code does the opposite (correctly, and the comment explains why). Also, the table’s percentages don’t match its own numbers (49s to 30s is about a 39% reduction, not 61%, and 49s to 28s is about 43%, not 77%). Mostly pointing this out for posterity, in case anyone ever looks back at this PR while researching or debugging a future issue.

One final thought: if `fork` raises `EAGAIN` partway through the fan-out, the already-spawned workers are never reaped and their readers leak. I agree with propagating the error, but an `ensure` in `fan_out` could still close readers and reap what was spawned before re-raising.

Happy to merge once the these issues are sorted.

sferik:
@danielwestendorf no pressure but I’m curious if you have a timeline for making these changes (or if you’d prefer I step-in to get this over the finish line). I was hoping to do a bugfix release next week and would love to include this change!

danielwestendorf:
@sferik yup, working on it now 👍

Issue #1262, opened 2026-08-13
Add a `patch` command that reports coverage of the lines a change touched


sferik:
`simplecov diff` compares two `coverage.json` files, which answers "did the number move" and needs a baseline artifact from an earlier run. The question a reviewer actually asks is a different one: is the code in this pull request tested? A project sitting at 40% cannot move its global number in one PR, but it can insist that every line it adds is covered.

`simplecov patch --base main` would read `git diff --unified=0 main...HEAD`, intersect the added and modified line numbers with the current report, and print coverage over only those lines:

```
$ simplecov patch --base main
 lib/simplecov/cli/patch.rb   88.00%  (22/25)   missing 41-43
 lib/simplecov/result.rb     100.00%  (4/4)
 Patch coverage: 89.66% (26/29)
```

`--minimum N` exits non-zero below a floor, so it composes as a CI gate alongside the existing threshold checks, and `--json` emits rows the way the other read-only commands do.

Most of what this needs already exists. The report knows which lines are relevant and which were hit, so the new machinery is a parser for unified diff hunk headers plus path resolution against `SimpleCov.root`. Renames want `--find-renames`, otherwise a moved file reads as entirely new and its untouched lines land in the denominator. Lines that `LinesClassifier` considers never relevant should stay out of that denominator too, the same way they stay out of the file total.

Prior art: `diff-cover` in Python, Codecov's patch status, and the `undercover` gem, which does this for Ruby but needs its own LCOV formatter and a separate git dependency to get there.

sferik:
Landed in https://github.com/simplecov-ruby/simplecov/pull/1280. Thanks @mswiszcz!

Issue #886, opened 2020-04-12
Support for "diff coverage"


deivid-rodriguez:
The only codeclimate feature that I think it's not provided by default in simplecov is "diff coverage", namely, the ability to track the coverage of a set of source control changes, and to enforce restrictions on it.

For example, check that a given patch cannot have less that 50% coverage, which would mean that at least half of the lines being added or changed would need to be hit by tests.

Not sure if adding support for this inside `simplecov` could couple it too much with source control systems. Maybe a first step that does not require any knowledge about `git` would be to, given a coverage report, allow to query information about the partial coverage of a given set of { file_name, line_number } duples. I think that would simplify implementing diff coverage inside a plugin or something like that.

Thoughts?

PragTob:
Hey @deivid-rodriguez !

Definitey it'd be a great feature - not sure I'd wanna solve it inside simplecov. 

I'll loop in @grodowski as he reached out to me some time ago about his [undercover](https://github.com/grodowski/undercover) which takes care of exactly that problem iirc.

I like your proposed interface as it takes care of a significant chunk of the problem to the best of my understanding while not relying on VCS information. The latter of which I really wouldn't wanna have in simplecov core - we have way too many integrations as is.

deivid-rodriguez:
Nice, I didn't know about `undercover`, I'll have a look at it!

PragTob:
@deivid-rodriguez are you happy with what you found? :)

deivid-rodriguez:
I never got to dig into this anymore :disappointed:

PR #1280, opened 2026-08-18
Add `simplecov patch` command for per-change coverage

mswiszcz:
Implements #1262.

## What

`simplecov patch` reports coverage over only the lines a change touched — the question `diff` doesn't answer. It reads `git diff --unified=0 <base>...HEAD`, intersects the added and modified line numbers with the current report (`--input coverage/coverage.json`), and prints line coverage — plus branch coverage over the branches those lines carry, when the report measured branches — for just that change.

```
$ simplecov patch --base main
   88.00% (22/25) lines   50.00% (1/2) branches  lib/simplecov/cli/patch.rb  missing 41-43  branch 39
  100.00% (4/4) lines  lib/simplecov/result.rb
  Patch coverage:  89.66% (26/29) lines, 50.00% (1/2) branches
```

Where `diff` compares two reports and asks whether the overall number moved — which a large or legacy project can't shift in one pull request — `patch` lets a project sitting at 40% still require that everything it *adds* is tested.

## Flags

- `--base REF` — ref to diff against (default `main`; in CI, pass the pull request's target branch, or its merge-base)
- `--minimum N` — exit non-zero below a floor; line coverage, and branch coverage when the report has it, must both clear it, so it gates a change alongside the existing thresholds even when those are already satisfied
- `--find-renames` — follow a moved file instead of counting it as entirely new
- `--json` — emit the rows as a JSON array
- `--input PATH` / `--no-color` — as with the other read-only subcommands

Only files the report already tracks are scored — a changed file outside the configured `cover` / `track_files` set is out of scope — and a touched line `LinesClassifier` deems never relevant (blank or comment) stays out of the denominator, so a comment-only change reports nothing to cover rather than a gap.

## On shelling out to git

#886 was closed partly over keeping a VCS dependency out of core. That constraint has since relaxed in practice: `JSONFormatter`'s `result_hash_formatter.rb#git_commit` already shells out to `git`. `patch` uses the same `Open3` + `git` approach and adds no gem dependency; it degrades cleanly when git isn't present or the ref doesn't resolve (reports on stderr and exits non-zero). Prior art: `diff-cover`, Codecov's patch status, and the `undercover` gem (Ruby-specific, but needs its own formatter and a git dependency to get there).

## Tests & docs

- 22 examples in `spec/cli_spec.rb` covering valid, edge, and invalid cases: renames, whole-file and pure deletions, malformed coverage entries, off-change and malformed branches, JSON output, `--minimum` gating on both criteria, and colorization.
- RBS signatures under `sig/internal/simplecov/cli/patch.rbs`; `rake steep` clean.
- Self-coverage stays at 100% line + branch + method (the dogfood gate).
- `CHANGELOG.md` and `docs/CLI.md` updated.


lib/simplecov/cli/patch.rb
```diff
@@ -0,0 +1,355 @@
...
+    # read-only commands do.
+    #
+    # Only files the report already carries are scored: a changed file
+    # SimpleCov never tracked (a non-Ruby file, or one outside the
+    # configured `cover` / `track_files` set) is out of scope, and a line
+    # `LinesClassifier` considers never relevant (blank or comment) stays
+    # out of the denominator the same way it stays out of the file total.
+    #
+    # rubocop:disable Metrics/ModuleLength -- one cohesive command: diff
+    # parsing, coverage intersection, and rendering read top-to-bottom here.
```
sferik: RuboCop’s default length limits are good rules of thumb, so I’d prefer your extract some of this code instead of disabling it. One natural seam is the git half: `git_diff` through `diff_path` as something like `Patch::ChangedLines`, which would give the diff parser its own focused file and drop this module under the limit.

lib/simplecov/cli/patch.rb
```diff
@@ -0,0 +1,355 @@
...
+        changes = changed_lines(opts[:base], find_renames: opts[:find_renames], stderr: stderr)
+        return 1 unless changes
+
+        rows = compute_rows(opts[:coverage], changes)
+        emit(stdout, rows, opts)
+        gate(rows, opts[:minimum])
+      end
+
+      def parse(args, stderr)
+        opts, = parse_common(args, base: DEFAULT_BASE, find_renames: false, minimum: nil) do |parser, options|
```
sferik: `opts, =` drops the positional arguments, so `simplecov patch feature-x` (a forgotten `--base`) silently diffs against `main`. For this command a stray positional looks exactly like a ref, which makes the silence costly: the gate can pass while measuring the wrong change. Erroring on a non-empty rest seems worth the two lines.

lib/simplecov/cli/patch.rb
```diff
@@ -0,0 +1,355 @@
...
+        # A git ref can never begin with "-", so refusing one keeps a
+        # `--base` value from being read by git as an option instead of a
+        # revision (e.g. `--output=FILE` writes to disk, `--line-prefix=`
+        # empties the diff so a `--minimum` gate passes over the change).
+        return nil if base.start_with?("-")
+
+        cmd = ["git", "-c", "core.quotePath=false", "diff", "--unified=0", "--relative",
+               "--no-color", "--no-ext-diff", "--no-textconv", "--inter-hunk-context=0",
+               "--src-prefix=a/", "--dst-prefix=b/"]
+        cmd += [find_renames ? "--find-renames" : "--no-renames", "#{base}...HEAD", "--"]
```
sferik: `<base>...HEAD` diffs the merge base against the HEAD commit, so uncommitted work is invisible: someone who writes code, runs the suite, and runs `simplecov patch` before committing sees "no coverable lines changed", and a `--minimum` gate passes vacuously. That is the right semantics for CI but not for local development. Consider `git diff --merge-base <base>`, which keeps the merge-base semantics and includes the working tree.

lib/simplecov/cli/patch.rb
```diff
@@ -0,0 +1,355 @@
...
+      # 100 and pass it against `--minimum 100`, while dividing
+      # (`covered / relevant * 100`) makes 23/40 compute as 57.4999… and
+      # fail an exactly-57.5% patch. `covered * 100` is exact and a
+      # representable minimum times an integer is too. A criterion with
+      # nothing relevant never falls short.
+      def short?(stats, minimum)
+        relevant = stats[:relevant]
+        return false if relevant.zero?
+
+        stats[:covered] * 100 < minimum * relevant
```
sferik: Cross-multiplying fixes the two failure modes the comment names, but it introduces a third: most decimal minimums are not representable in binary, so `minimum * relevant` can land a hair above the exact product. Concretely, a patch at exactly 64.4% (161 of 250 lines) against `--minimum 64.4` computes `16100 < 64.4 * 250`, which is `16100 < 16100.000000000002`, and fails a change that exactly meets the floor. Going one step further in the same spirit makes the gate exact:

```suggestion
        stats[:covered] * 100 < Rational(minimum.to_s) * relevant
```

sferik (changes requested):
This looks very good overall! It fits the CLI architecture and the `CoverageFile.lookup` rewrite looks particularly good.

I’ve noted a few minor issues but it should be relatively straightforward to get this PR merged. The branch also needs a rebase, since the `CHANGELOG` has changed since you branched.

Thanks for this patch!

sferik (approved):
Much better now. Thank you!

Issue #1299, opened 2026-09-22
Load error with JRuby on Windows with the latest version


kiskoza:
**Environment Information**
  
Provide at least:
* Operating system and platform (e.g. `uname -a`)
```
Operating System
  Microsoft Windows Server 2025
  10.0.26100
  Datacenter
Runner Image
  Image: windows-2025-vs2026
  Version: 20260907.229.1
  Included Software: https://github.com/actions/runner-images/blob/win25-vs2026/20260907.229/images/windows/Windows2025-VS2026-Readme.md
  Image Release: https://github.com/actions/runner-images/releases/tag/win25-vs2026%2F20260907.229
```

* JRuby version (`jruby -v`) and command line (flags, JRUBY_OPTS, etc)
```
C:\Windows\system32\cmd.exe /D /S /C "D:\jruby-10.0.6.0\bin\ruby.bat --version"
  jruby 10.0.6.0 (3.4.5) 2026-06-11 716ad51f54 OpenJDK 64-Bit Server VM 21.0.12.1+1-LTS on 21.0.12.1+1-LTS +indy +jit [x86_64-mswin32]
```

Other relevant info you may wish to add:
* I'm trying to run code on the CI with Github Runners
* I checked the list of pre-installed software on the CI runner images, but relevant parts (e.g.: Java version) seems to be the same.

**Expected Behavior**

I have a CI for the CAxlsx gem that is testing JRuby as part of the test matrix. It used to pass with previous versions ( [example pass](https://github.com/caxlsx/caxlsx/actions/runs/34105594153/job/101689746968) ), but recently it started to fail ( [example fail](https://github.com/caxlsx/caxlsx/actions/runs/35342414679/job/106386937755?pr=546) ). As far as I know there were no new prism or JRuby releases between the failing and passing CIs.

**Actual Behavior**

The Ruby install seems to be working fine, up until a point where it fails on loading libprism.dll when running rake

```
LoadError: Could not open library 'D:/jruby-10.0.6.0/lib/prism/libprism.dll' : The specified procedure could not be found.
.
Searched in <system library path>
            load_library at D:/jruby-10.0.6.0/lib/ruby/stdlib/ffi/dynamic_library.rb:94
                 ffi_lib at D:/jruby-10.0.6.0/lib/ruby/stdlib/ffi/library.rb:95
                     map at org/jruby/RubyArray.java:2896
[...]

PR #1305, opened 2026-09-22
Look up Action View handlers on the new registry

### LINKED ISSUE #1300: `cover_views` broken on rails main
You use `registered_template_handler` (a private API) which rails refactored away in https://github.com/rails/rails/commit/cd416503e86239f2365d0973f045caca8ac2a764. When running against rails main, you end up with a bunch of warnings like this:
> [SimpleCov]: Skipping .../app/views/application/404.html.erb, which did not compile (NoMethodError)

Removing the rescue confirms the error:
```
/home/user/code/simplecov/lib/simplecov/view_coverage/template_compiler.rb:72:in 'SimpleCov::ViewCoverage::TemplateCompiler#build_template': undefined method 'registered_template_handler' for class ActionView::Template (NoMethodError)

        handler = ActionView::Template.registered_template_handler(extension)
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Did you mean?  register_template_handler
               unregister_template_handler
        from /home/user/code/simplecov/lib/simplecov/view_coverage/template_compiler.rb:36:in 'SimpleCov::ViewCoverage::TemplateCompiler#call'
        from /home/user/code/simplecov/lib/simplecov/view_coverage.rb:32:in 'block in SimpleCov::ViewCoverage#compile_unrendered'
```

Saidbek:
Fixes #1300.

`cover_views` called `ActionView::Template.registered_template_handler`, which Rails removed in https://github.com/rails/rails/pull/57874 when the handler registry moved onto `ActionView::Template::Handlers`. Every unrendered template was skipped with `NoMethodError`.

The lookup now reads `ActionView::Template::Handlers.template_handlers` on current Rails and keeps the old method when it still exists. An extension with no handler is still skipped. `handler_for_extension` is not used, because it falls back to the raw handler and would report those files as static text.


sferik (approved):
Thank you for reporting this issue and submitting this fix. Looks good to me!

This patch will ship in version 1.3.1, which I will ship as soon as I have a resolution to https://github.com/simplecov-ruby/simplecov/issues/1299.

Thanks again!

sferik:
The fix for this bug should be live in version 1.3.1. I’d appreciate if you could either confirm the fix or reopen https://github.com/simplecov-ruby/simplecov/issues/1300.

